PR #3, opened 2019-12-23
Handle review requests

lucasocon:
Generate a webhook controller to receive the posts made to the endpoint /event_handler, it calls the service GitHub handler and executes the correct method based on the event type header and an action. 
For now, the only event supported is the review request event under the 'pull request' event type.

`handle_review_request` finds the generating user, a.k.a review_request's owner, and the participants, or creates them if they do not exist (based on their github_id). 

Likewise with the PR. it then relates the models in the creation step. It saves the JSON string as an attribute 'data' as well.
A GitHub app needs to be created and its webhook needs to be set to:
the machine's IP address (or a ngrok tunnel) + /event_handler in the URL


app/services/github_handler.rb
```diff
@@ -0,0 +1,29 @@
...
+      # else unsupported
+    end
+  end
+
+  def handle_review_request
+    owner = User.create_or_find(@payload["pull_request"]["user"])
+    pr = PullRequest.create_or_find(@payload["pull_request"])
+    # Even if you select multiple reviewers at once, the webhook sends a post for every person selected
+    # we can assume it is going to be a single user obj "requested_reviewer"
+    participants = [ User.create_or_find(@payload["requested_reviewer"]) ]
```
santib: https://github.com/rootstrap/tech-guides/tree/master/ruby#spaces-braces

.gitignore
```diff
@@ -32,3 +32,7 @@
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
+
+#ignore rubocop files
+.rubocop.yml
```
santib: this file shouldn't be ignored

app/models/review_request.rb
```diff
@@ -0,0 +1,5 @@
+class ReviewRequest < ApplicationRecord
+  belongs_to :owner, class_name: "User", foreign_key: "owner_id", inverse_of: :owned_review_requests
```
santib: minor comment: you can use a symbol for the `owner_id`

app/models/review_request.rb
```diff
@@ -0,0 +1,5 @@
...
+  belongs_to :pull_request, inverse_of: :review_requests
+  has_and_belongs_to_many :users
```
santib: A personal opinion, I'd always go with `has_many :through` because it's more versatile

app/services/github_handler.rb
```diff
@@ -0,0 +1,29 @@
+class GithubHandler
+  def initialize (event, payload)
+    @event = event
+    @payload = payload
+  end
+
+  def handle
```
santib: are we validating that the request comes actually from github? is there a ticket to add that in the future?

app/models/user.rb
```diff
@@ -0,0 +1,9 @@
+class User < ApplicationRecord
+  has_and_belongs_to_many :review_requests
+  has_many :owned_review_requests, class_name: "ReviewRequest", foreign_key: "owner_id", dependent: :destroy, inverse_of: :owner
+
+  def self.create_or_find (json)
```
santib: https://github.com/rootstrap/tech-guides/tree/master/ruby#spaces-braces

santib: Honestly, I don't think the model should know about a json structure coming from Github. I'd rather leave this code in the service. What do you think?

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
+  post '/event_handler', to: 'webhook#handle'
```
santib: will this route be just for github or could it be by something else in the future? asking because if it's tightly coupled to github then I'd make it clear in the route

db/migrate/20191220183838_create_users.rb
```diff
@@ -0,0 +1,11 @@
+class CreateUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :users do |t|
+      t.string :login
```
santib: is any of these attributes required? asking because maybe we can add a non-null constraint

db/migrate/20191220183808_create_review_requests.rb
```diff
@@ -0,0 +1,9 @@
+class CreateReviewRequests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :review_requests do |t|
+      t.text :data
```
santib: just asking: is this a text or should this be a `jsonb`?

db/migrate/20191220190613_create_review_request_users_join_table.rb
```diff
@@ -0,0 +1,8 @@
+class CreateReviewRequestUsersJoinTable < ActiveRecord::Migration[6.0]
+  def change
+    create_join_table :review_requests, :users do |t|
+      t.index [:review_request_id, :user_id]
+      # t.index [:user_id, :review_request_id]
```
santib: remove unnecessary comment

app/models/user.rb
santib: this makes me think.. is rubocop correctly configured? Why didn't it catch any of these syntax errors?

db/migrate/20191223133115_create_pull_requests.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePullRequests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :pull_requests do |t|
+      t.integer :github_id
+      t.integer :number
+      t.string :state
+      t.boolean :locked
```
santib: avoid the three state boolean problem https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem

db/migrate/20191223133115_create_pull_requests.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePullRequests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :pull_requests do |t|
+      t.integer :github_id
```
santib: are you sure an integer is enough? maybe we need a `bigint`?

app/services/github_handler.rb
```diff
@@ -0,0 +1,29 @@
+class GithubHandler
```
santib: Would it be possible to add some tests?

.gitignore
```diff
@@ -32,3 +32,7 @@
...
+.rubocop_todo.yml
```
santib: what is this file? If you need to ignore files locally you can add them to your personal `~/.gitignore` (home directory)

.env.example
```diff
@@ -1,3 +1,6 @@
 # We need to set both the username and password as environment variables in order to use the app
 GITHUB_ANALYZER_USERNAME=postgres
-GITHUB_ANALYZER_PASSWORD=postgres
\ No newline at end of file
+GITHUB_ANALYZER_PASSWORD=postgres
+
+# Must be set as the webhook secret on your github app
+GITHUB_ANALYZER_WEBHOOK_SECRET=78e59601e29b4db895f86fe6ac4442da9a6b8843
```
santib: 🤔 don't commit actual keys. you can have `fake-key` as the value in the `.env.example`

.rubocop.yml
```diff
@@ -0,0 +1,110 @@
...
+Lint/AmbiguousBlockAssociation:
+  Exclude:
+    - spec/**/*
+
+Metrics/AbcSize:
+  # The ABC size is a calculated magnitude, so this number can be an Integer or
+  # a Float.
+  Max: 15
+  Exclude:
+    - db/migrate/20161011151353_devise_create_users.rb
```
santiagovidal: This doesn't exist in this project.

.rubocop.yml
```diff
@@ -0,0 +1,110 @@
...
+  CountComments: false  # count full line comments?
+  Max: 200
+
+Metrics/LineLength:
+  Max: 100
+  # To make it possible to copy or click on URIs in the code, we allow lines
+  # containing a URI to be longer than Max.
+  AllowURI: true
+  Exclude:
+    - db/migrate/20161011184702_devise_create_admin_users.rb
```
santiagovidal: This and the one below either

.rubocop.yml
```diff
@@ -0,0 +1,110 @@
...
+Metrics/PerceivedComplexity:
+  Max: 12
+
+Rails:
+  Enabled: true
+
+Rails/Delegate:
+  Enabled: true
+  Exclude:
+    - app/controllers/api/concerns/act_as_api_request.rb
```
santiagovidal: Or this

Gemfile
```diff
@@ -28,22 +28,22 @@ gem "pg", "~> 1.1"
 gem 'bootsnap', '>= 1.4.2', require: false
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
-gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
-gem "github_api"
+gem 'github_api'
```
gonzaM: I would add the version here.

app/models/review_request.rb
gonzaM: I agree with Santi. I would use `has_many :through` here and below.

app/controllers/webhook_controller.rb
```diff
@@ -0,0 +1,13 @@
+class WebhookController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  def handle
+    @event = request.headers['X-GitHub-Event']
+    @signature = request.headers['X-Hub-Signature']
+    # this is needed because the action attr is replaced by rails' action param
+    # and is lost on the params hash
+    @payload = request.raw_post
```
gonzaM: Are the `@` necessary?

app/services/github_handler.rb
```diff
@@ -0,0 +1,64 @@
...
+  def initialize(event, payload, signature)
+    @event = event
+    @payload = JSON.parse(payload)
+    @raw_payload = payload
+    @signature = signature
+  end
```
gonzaM: I would use `attr_reader` for payload, etc.

app/services/github_handler.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+
+  def handle_review_request
+    owner = create_or_find_user(@payload["pull_request"]["user"])
+    pr = create_or_find_pr(@payload["pull_request"])
+    # Even if you select multiple reviewers at once, the webhook sends a post for every person selected
+    # we can assume it is going to be a single user obj "requested_reviewer"
+    participants = [create_or_find_user(@payload["requested_reviewer"])]
+    ReviewRequest.create(data: @payload, owner: owner, users: participants, pull_request: pr)
```
gonzaM: I would use something like `pr.review_requests.create(data: @payload, owner: owner, users: participants)`

gonzaM: What happens if this creation fails? (`create` vs `create!`?)

db/migrate/20191226185853_add_not_null_booleans_to_pull_request.rb
```diff
@@ -0,0 +1,8 @@
+class AddNotNullBooleansToPullRequest < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :pull_requests, :locked, false, false
+    change_column_null :pull_requests, :merged, false, false
+    change_column_null :pull_requests, :draft, false, false
+    change_column_null :pull_requests, :locked, false, false
+  end
```
gonzaM: Could you add the validations in the model too?

app/services/github_handler.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def webhook_verified?
+    digest = OpenSSL::HMAC.hexdigest("SHA1", ENV["GITHUB_ANALYZER_WEBHOOK_SECRET"], @raw_payload)
+    ActiveSupport::SecurityUtils.secure_compare("sha1=#{digest}", @signature)
+  end
+
+  def create_or_find_user(json)
+    User.create_or_find_by(
```
santib: same here for the usage of `!`

santib: if the insertion fails because of a constraint, then it will try to find by all of the attributes. What happens if the `github_id` already exists in the db but the record has a different `login` ?

db/migrate/20191223133115_create_pull_requests.rb
santib: I think this should be not null in all of the tables

app/models/review_request.rb
lucasocon: I had initially thought that a review request is related to multiple users due to the array requested_reviewers in the response, but it turns out it sends a unique requested_reviewer per request. I think it's better to relate it to a single user and a pull request, to simulate the request better, and remove the habtm relation entirely. What do you think?

santib: makes sense

app/services/github_handler.rb
```diff
@@ -0,0 +1,84 @@
...
+        node_id: json['node_id'],
+        login: json['login'],
+        github_id: json['id']
+      )
+    end
+    user
+  end
+
+  def create_or_find_pr(json)
+    pr = PullRequest.find_by(json['github_id'])
```
GuilleLeopold: I think you are missing the attribute that you are using to `find_by`, `find_by(github_id: json['github_id'])` right?

app/services/github_handler.rb
```diff
@@ -0,0 +1,84 @@
...
+        number: json['number'],
+        state: json['state'],
+        locked: json['locked'],
+        draft: json['draft'],
+        title: json['title'],
+        body: json['body'],
+        closed_at: json['closed_at'],
+        merged_at: json['merged_at'],
+        merged: json['merged'],
+        github_id: json['id']
```
GuilleLeopold: shouldn't be github_id? or the one from above just id? (In the json)

app/services/github_handler.rb
```diff
@@ -0,0 +1,84 @@
...
+        login: json['login'],
+        github_id: json['id']
+      )
+    end
+    user
+  end
+
+  def create_or_find_pr(json)
+    pr = PullRequest.find_by(json['github_id'])
+    if pr.nil?
```
GuilleLeopold: Also I think you can just do something like `PullRequest.find_by(...) || PullRequest.create!(....)`

spec/factories/users.rb
```diff
@@ -0,0 +1,23 @@
...
+# Indexes
+#
+#  index_users_on_github_id  (github_id) UNIQUE
+#
+
+FactoryBot.define do
+  factory :user do
+    sequence(:github_id, 1000)
+    sequence(:login, 100) { |n| "octocat#{n}" }
+    node_id { 'MDQ6VXNlcjE=' }
```
GuilleLeopold: No need for brackets here as it's an atomic attribute

spec/models/pull_request_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      pr.merged = nil
+      expect(pr.valid?).to eq(false)
+    end
+
+    it 'does not allow draft to be null' do
+      pr.draft = nil
+      expect(pr.valid?).to eq(false)
+    end
+
+    it 'should have a unique github_id' do
```
GuilleLeopold: `should` is depreated, just `has a unique github_id` it's ok

spec/models/pull_request_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      FactoryBot.create(:pull_request, github_id: '0001')
```
GuilleLeopold: Just `create(:pull_request, github_id: '0001') works fine, it already calls to factorybot, the same below

spec/factories/users.rb
santib: I think they got deprecated https://thoughtbot.com/blog/deprecating-static-attributes-in-factory_bot-4-11

app/services/github_handler.rb
```diff
@@ -0,0 +1,84 @@
...
+    @event = event
+    @payload = JSON.parse(payload)
+    @raw_payload = payload
+    @signature = signature
+  end
+
+  def handle
+    return FORBIDDEN unless webhook_verified?
+
+    case event
```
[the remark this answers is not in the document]
lucasocon: Other events will be tracked as well

PR #4, opened 2020-01-21
Handling review request removed

hdamico:
## Description

Now we can handle `review_request_removed` requests from github.


app/services/github_handler.rb
```diff
@@ -45,6 +45,14 @@ def handle_review_request
     pr.review_requests.create!(data: payload, owner: owner, reviewer: reviewer)
   end
 
+  def handle_review_removal
+    pr = PullRequest.find_by!(github_id: payload['pull_request']['id'])
+    pr.transaction do
+      pr.review_requests = []
```
santib: why is this needed? Isn't `dependent: :destroy` here https://github.com/rootstrap/rs-code-review-metrics/blob/master/app/models/pull_request.rb#L26 going to take care of this already?

hdamico: You are right! Committing the changes

app/services/github_handler.rb
```diff
@@ -45,6 +45,10 @@ def handle_review_request
     pr.review_requests.create!(data: payload, owner: owner, reviewer: reviewer)
   end
 
+  def handle_review_removal
+    PullRequest.find_by!(github_id: payload['pull_request']['id']).destroy!
```
santib: Maybe I'm missing something, but why removing a review request destroys a `PullRequest`? I thought we would be just destroying the `ReviewRequest`

hdamico: Yeah, i got a miss understanding. Thanks!

app/services/github_handler.rb
```diff
@@ -45,6 +45,12 @@ def handle_review_request
     pr.review_requests.create!(data: payload, owner: owner, reviewer: reviewer)
   end
 
+  def handle_review_removal
+    reviewer = User.find_by(github_id: payload['requested_reviewer']['id'])
+    pr = PullRequest.find_by(github_id: payload['pull_request']['id'])
+    pr.review_requests.find_by!(reviewer: reviewer).destroy!
```
santib: do we care about an exception if the review_request isn't found?

maybe we can do `pr.review_requests.find_by(reviewer: reviewer)&.destroy!`

thougths?

spec/services/github_handler_spec.rb
```diff
@@ -81,5 +81,22 @@
...
+        create :user,
+          github_id: 1001,
+          login: 'pentacat',
+          node_id: 'MDExOlB1bGxc5MTQ3NDM3'
+      }
+      let!(:review_requests) { ReviewRequest.create(
+                                pull_request: pull_request, owner: user, reviewer: user
+                              ) }
+
+      it 'deletes review requests' do
```
santib: minor comment, `deleted review request` in singular

app/services/github_handler.rb
santiagovidal: Do we want to destroy these requests? Or mark them as invalid?

For example, if person A sends a review request to person B and then removes it after 3 days, do we want that review request to affect the statistics?

I think so but what does everyone think?

@anthfb @matiasmansilla1989 @santib

spec/services/github_handler_spec.rb
```diff
@@ -81,5 +81,22 @@
...
+
+    describe 'review removal request' do
+      let!(:pull_request) { create :pull_request, github_id: 1000 }
+      let!(:user) {
+        create :user,
+          github_id: 1001,
+          login: 'pentacat',
+          node_id: 'MDExOlB1bGxc5MTQ3NDM3'
+      }
+      let!(:review_requests) { ReviewRequest.create(
```
santiagovidal: `do ... end` blocks are preferred for multi-line `let` variables. Our linters will report this once we add them to the project.

db/migrate/20200122143238_add_status_to_review_request.rb
```diff
@@ -0,0 +1,5 @@
+class AddStatusToReviewRequest < ActiveRecord::Migration[6.0]
+  def change
+    add_column :review_requests, :status, :string
```
santib: shouldn't we use an `enum` for statuses?

hdamico: Done!

app/models/review_request.rb
```diff
@@ -31,4 +31,11 @@ class ReviewRequest < ApplicationRecord
              foreign_key: :reviewer_id,
              inverse_of: :received_review_requests
   belongs_to :pull_request, inverse_of: :review_requests
+
+  enum status: { active: 'active', removed: 'removed' }
+  validates :status, inclusion: { in: statuses.keys }
+
+  def remove
```
santib: you don't need this method, rails enums already gives you some handful method, for example in your case you want to do `pr.review_requests.find_by!(reviewer: reviewer)&.removed!`

PR #5, opened 2020-01-21
Add CODEOWNERS file to check pull request payload

hdamico:
## Description

This PR adds a test CODEOWNER file to check if `opened` action from `pull_request` event receives the code owners, or if it generates separated requests.

## THIS MUST BE DELETED

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*.js    @hdamico


CODEOWNERS
```diff
@@ -0,0 +1 @@
+*.js    @hdamico
```
horacio: You can leave just `@hdamico` (to be a global code owner) or change this line for now to `*.rb`. WDYT?

PR #14, opened 2020-01-22
Handle open and closed requests

### LINKED ISSUE #10: Handle 'opened' and 'closed' requests
## Description

Handle 'opened' as a Create and 'closed' as a Destroy of Pull Requests, but do not delete them from the database, just set the status to 'closed'

## Things to do

* Make a refactor of `find_or_create`
* Catch `opened` and `closed` actions

hdamico:
## Background

Resolves #10

### DIFF
--- a/app/services/github_handler.rb
+++ b/app/services/github_handler.rb
@@ -9,8 +9,8 @@ class GithubHandler
 
   def initialize(event, payload, signature)
     @event = event
-    @payload = JSON.parse(payload)
     @raw_payload = payload
+    @payload = parse_payload
     @signature = signature
   end
 
@@ -19,11 +19,15 @@ class GithubHandler
 
     case event
     when 'pull_request'
-      case payload['action']
+      case payload.action
+      when 'opened'
+        opened
       when 'review_requested'
-        handle_review_request
+        review_request
+      when 'closed'
+        closed
       when 'review_request_removed'
-        handle_review_removal
+        review_removal
       else
         return BAD_REQUEST
       end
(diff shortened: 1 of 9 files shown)


app/services/github_handler.rb
```diff
@@ -35,50 +39,59 @@ def handle
...
+      draft: pr_data.draft,
+      title: pr_data.title,
+      body: pr_data.body,
+      closed_at: pr_data.closed_at,
+      merged_at: pr_data.merged_at,
+      merged: pr_data.merged,
+      github_id: pr_data.id
+    )
+  rescue ActiveRecord::RecordNotUnique => e
+    errors.add(:base, 'This pull request already exists!')
```
santib: what is this error for? the request is triggered by a webhook

spec/services/github_handler_spec.rb
```diff
@@ -48,23 +49,35 @@
...
+    describe 'pull request close' do
+      let!(:pull_request) { create :pull_request, github_id: 1000 }
+
+      before do
+        github_handler.closed
+        pull_request.reload
+      end
+
+      it 'sets status closed' do
+        expect(pull_request.closed?).to be true
```
santib: `expect { github_handler.closed }.to change { pull_request.reload.closed? }.from(false).to(true)`

spec/services/github_handler_spec.rb
```diff
@@ -48,23 +49,35 @@
...
+          github_handler.create_or_find_user(payload.requested_reviewer)
         }.to change(User, :count).by(0)
       end
     end
 
     describe 'pull request not existing' do
       it 'adds pr to the DB' do
         expect {
-          github_handler.create_or_find_pr(payload['pull_request'])
+          github_handler.opened
```
santib: I'd reorganize the test file like

`describe '#opened' do`
`describe '#closed' do`
`describe '#review_requested' do`
`describe '#review_removal' do`

http://www.betterspecs.org/#describe

app/services/github_handler.rb
hdamico: Do you mean to delete the rescue? I made this for future logs, but I guess it's ok to have the exception on the logs

santib: where is `errors` defined? if you wanted to log something maybe you meant `Rails.logger.error('This pull request already exists!')` ?

hdamico: Yeah! I mean to do that

app/services/github_handler.rb
```diff
@@ -35,50 +39,59 @@ def handle
...
+      draft: pr_data.draft,
+      title: pr_data.title,
+      body: pr_data.body,
+      closed_at: pr_data.closed_at,
+      merged_at: pr_data.merged_at,
+      merged: pr_data.merged,
+      github_id: pr_data.id
+    )
+  rescue ActiveRecord::RecordNotUnique
+    Rails.logger.error('This pull request already exists!')
```
santib: As I mentioned in the previous comment I think it's more correct to just let the exception happen if we are not going to handle it. This log will even give us less information on what happened than the exception

hdamico: Ok!

PR #16, opened 2020-01-23
Sidekiq implemented

### LINKED ISSUE #9: Implement Sidekiq
## Description

We need to process every request in the background.

## Things to do

* Implement Sidekiq
* Create Workers

hdamico:
## What does this PR do?

Adds Sidekiq support.

Resolves #9


.rubocop.yml
```diff
@@ -1,100 +0,0 @@
-require: rubocop-rails
```
santib: this file shouldn't be deleted

Gemfile
```diff
@@ -33,6 +33,8 @@ gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 gem 'github_api', '~> 0.18'
 
 gem 'annotate', '~> 3.0'
+# Use Sidekiq to run background threads
+gem 'sidekiq'
```
santib: specify gem version

app/helpers/payload_parser.rb
```diff
@@ -0,0 +1,5 @@
+module PayloadParser
```
santib: `app/helpers` directory has another use in rails, it's actually view helpers

.rubocop.yml
```diff
@@ -1,100 +1,100 @@
 require: rubocop-rails
 
-AllCops:
```
santib: the **file** shouldn't be deleted

.rubocop.yml
```diff
@@ -1,100 +1,100 @@
...
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
-Style/ModuleFunction:
-  Enabled: false
-
-Style/ReturnNil:
-  Enabled: true
\ No newline at end of file
+AllCops:
```
santib: don't add unnecessary whitespaces

app/services/github_handler.rb
```diff
@@ -45,51 +47,18 @@ def webhook_verified?
...
-      locked: pr_data.locked,
-      draft: pr_data.draft,
-      title: pr_data.title,
-      body: pr_data.body,
-      closed_at: pr_data.closed_at,
-      merged_at: pr_data.merged_at,
-      merged: pr_data.merged,
-      github_id: pr_data.id
-    )
+    GithubService::OpenedWorker.perform_async(raw_payload)
```
santib: use activejob on top of sidekiq

app/workers/github_service/closed_worker.rb
```diff
@@ -0,0 +1,11 @@
+module GithubService
+  class ClosedWorker
+    include Sidekiq::Worker
+    include PayloadParser
+
+    def perform(payload)
+      payload = parse_payload(payload)
```
santib: I'd rather have jobs not have logic, and just call `services`

spec/services/github_handler_spec.rb
```diff
@@ -36,29 +36,6 @@
       it 'creates a review request' do
         expect { github_handler.review_request }.to change(ReviewRequest, :count).by(1)
       end
-
-      it 'adds user to the DB' do
```
santib: we don't want to remove tests

app/helpers/payload_parser.rb
hdamico: `app/services/concerns` is ok?

santib: hmmm concerns has a particular meaning for rails (which I dislike) somethine like `app/lib`?

app/workers/github_service/closed_worker.rb
hdamico: What do you mean? Keep the method in the `GithubHandler` service and then call it from the job? Like `GithubHandler.closed` ?

santib: I mean having POROs where we have the business logic and the job just calls it. for example
```
# app/jobs/github/closed_job.rb
module Github
  class ClosedJob
    ...

    def perform(payload)
       Github::ClosedService.new(payload).call
    end
  end
end

# app/services/github/closed_service.rb
module Github
  class ClosedService
    attr_reader :payload

    def initialize(payload)
      @payload = payload
    end

    def call
      # logic goes here
    end
  end
end
```

A related link about jobs: https://chriskottom.com/blog/2015/11/bulletproof-rails-background-jobs/
A related link about services and more: https://codeclimate.com/blog/7-ways-to-decompose-fat-activerecord-models/

app/jobs/github_jobs/closed_job.rb
```diff
@@ -0,0 +1,9 @@
+module GithubJobs
+  class ClosedJob < ApplicationJob
+    queue_as :closed
```
santib: does this queue even exist? By default sidekiq uses just the `default` queue. BTW it's not recommended by sidekiq to have lot of queues, I think `default` `high` and `low` could be enough for now.  https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues

app/services/base_service.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+ 
+# Base class for service classes
+class BaseService 
+    private_class_method :new
```
santib: extra spaces

app/services/github_handler.rb
```diff
@@ -1,16 +1,16 @@
 class GithubHandler
+
```
santib: unnecessary newline

app/services/github_handler.rb
```diff
@@ -40,56 +40,23 @@ def handle
...
-      merged_at: pr_data.merged_at,
-      merged: pr_data.merged,
-      github_id: pr_data.id
-    )
+    GithubJobs::OpenedJob.perform_later(payload)
   end
 
   def closed
-    find_pr.closed!
+    GithubJobs::ClosedJob.perform_later(payload)
```
santib: how big can these payloads be? Isn't it a problem for Redis?

config/application.rb
```diff
@@ -16,6 +16,11 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    # Generates a "structure.sql" and allow raw sql
     config.active_record.schema_format = :sql
+    # ActiveJob abstraction
+    config.active_job.queue_adapter = :sidekiq
```
santib: maybe this can go just in production env file? For development I'm ok using sidekiq but we could choose to not use it. And for testing we should use `config.active_job.queue_adapter = :test`

spec/support/sidekiq.rb
```diff
@@ -0,0 +1,3 @@
+require 'sidekiq/testing'
```
santib: this file is not necessary

app/services/github_handler.rb
hdamico: There aren't big, I don't think it would be a problem for the future

app/services/base_service.rb
santib: the indentation is still wrong (4 spaces instead of 2). doesn't rubocop compain?

spec/support/sidekiq.rb
hdamico: This is necessary to test the jobs, otherwise they will run in the background.

santib: it's not if you use Rails config `config.active_job.queue_adapter = :test` https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/TestAdapter.html

PR #20, opened 2020-01-24
Setup Github Actions

### LINKED ISSUE #13: Integrate Github Actions (CI)

hdamico:
## What does this PR do?

Setting up Github Action for Ruby. Executes Rspec.

Resolves #13

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,65 @@
+name: Main
+
+on: [push]
+
+jobs:
+  vm-job:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,65 @@
+name: Main
```
santib: can we add to the workflow (or have another one?) for rubocop? 

example: https://github.com/rails/rails/actions?query=workflow%3ARuboCop

hdamico: Of course, we might have to create a list of functions.

PR #30, opened 2020-01-29
Event model added

### LINKED ISSUE #29: Create Event model
## Description

Generate `Event` model to catch every event from Github even if we don't handle it yet.

## Things to Do

* Delete `data` column from `ReviewRequest`
* Add `Event` model with `name` and `data`
* Create corresponding tests
* Modify `github_handler` service to call `Event` model in the different requests

hdamico:
## What does this PR do?

This PR changes most of the structure that we were using.
Now we have `Event` model which will handle different events like `Pull Request`, so now the relation is, an `Event` belongs to a type of event and a type of event has many `Event`.
We wiped `GithubHandler` and leave model `Event` to choose which event should be called.
Changed specs to use `shoulda-matchers` on model validations.

## Things left to do

- [x] Specs
- [x] Merged action

Resolves #29


app/services/github_service.rb
```diff
@@ -6,12 +6,19 @@ def initialize(payload)
   end
 
   def closed
-    PullRequest.find_by!(github_id: @payload.pull_request.id).closed!
+    Events::PullRequest.find_by!(github_id: @payload.pull_request.id)
+                       .update!(closed_at: Time.now)
```
santiagovidal: `Time.current` is preferable to take the server's timezone into account.

app/models/event.rb
```diff
@@ -0,0 +1,57 @@
...
+  def resolve
+    return handle if handleable?
+
+    save!
+  end
+
+  private
+
+  def handle
+    Event.create!(build_attributes)
```
santiagovidal: Are we duplicating the `Event` here? 🤔

app/services/github_handler.rb
```diff
@@ -1,65 +1,18 @@
...
-    when 'review_requested'
-      review_request
-    when 'closed'
-      closed
-    when 'review_request_removed'
-      review_removal
-    else
-      BAD_REQUEST
-    end
+    Event.create!(data: payload, name: event).handle if webhook_verified?
```
santiagovidal: Isn't `handle` a private method?

app/models/event.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def build_attributes
+    {
+      handleable: resolve_event,
```
santiagovidal: Are we creating the `handleable` object here? I'm asking because the `opened` method in the GithubService already creates the PullRequest object.

app/models/event.rb
hdamico: If it is handleable it will create a `Event` with, for example, `PullRequest`, otherwise it will create a `Event` with just the name and payload.
Now we have one `PullRequest` which has multiple `Events`.

app/models/event.rb
hdamico: No, it will depend on the `action` that we are receiving, if the event has the `opened` action it will create a PR, otherwise it will find the PR and assign it there.

app/services/github_handler.rb
hdamico: Yep, is `resolve` no `handle`, thanks for the correction!

hdamico:
@santiagovidal @santib @horacio
Please give a detailed review! :)

Gemfile
```diff
@@ -63,5 +63,6 @@ group :test do
   gem 'selenium-webdriver'
   # Easy installation and use of web drivers to run system tests with browsers
   gem 'simplecov', require: false
+  gem 'shoulda-matchers', require: false
```
santiagovidal: It's a good practice to lock down a version.

app/models/events/pull_request.rb
```diff
@@ -0,0 +1,83 @@
...
+        PullRequestJobs::ReviewRequestedJob.perform_later(payload)
+      end
+
+      def review_request_removed(payload)
+        PullRequestJobs::ReviewRequestRemovedJob.perform_later(payload)
+      end
+
+      private
+
+      ACTIONS = %w[opened review_requested closed \
```
santiagovidal: According to our style guides (https://github.com/rootstrap/tech-guides/blob/master/ruby/rails.md), constants should go at the top of the class.

app/models/event.rb
```diff
@@ -0,0 +1,48 @@
...
+class Event < ApplicationRecord
+  belongs_to :handleable, polymorphic: true, optional: true
+  validates :name, :data, presence: true
+
+  class << self
+    def resolve(payload)
+      event = payload[:event]
+      return handle(payload) if handleable?(event)
+
+      EventJob.perform_later(payload, event)
```
santiagovidal: Aren't we creating the event twice? Here and in `assign_event`.

lib/event_associator.rb
```diff
@@ -0,0 +1,7 @@
+module EventAssociator
```
santiagovidal: I don't know if I would put this in `lib`. This has knowledge of the models, and therefore of the business logic, so I would put it in `app` (as a part of a service or something).

app/models/event.rb
hdamico: We don't create the Event until it reaches the event service, or if don't handle it

lib/event_associator.rb
hdamico: I guess i can put it in the concern folder, what do you think?

app/controllers/webhook_controller.rb
```diff
@@ -1,20 +1,22 @@
...
-      head :forbidden
-    when GithubHandler::BAD_REQUEST
-      head :bad_request
-    end
+    Event.resolve(@payload) if webhook_verified?
+  end
+
+  private
+
+  def format_params
```
horacio: I think we can come up with a better name for this. What we're really doing here is setting up the payload. Maybe we can continue with the handling of the event only if it's verified.

What do you think about something like...


```ruby
def set_payload
  event = request.headers['X-GitHub-Event']
  signature = request.headers['X-Hub-Signature']

  payload = JSON.parse(request.raw_post).merge(event: event)

  head 400 and return unless payload_verified?(payload, signature)
end

```

One more thing. Shouldn't `handle` return, for example, `head :ok`?

hdamico: Yes! I had it in that way i don't know why did i change it. The head :ok and head 400 looks good, adding...

app/models/concerns/event_associator.rb
```diff
@@ -0,0 +1,7 @@
+module EventAssociator
+  def assign_event(payload, event)
```
santiagovidal: What if we add this method to `GithubService`? That way it's inherited by the different services and we don't have to add a new concern here.

spec/models/event_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+# Indexes
+#
+#  index_events_on_handleable_type_and_handleable_id  (handleable_type,handleable_id)
+#
+
+require 'rails_helper'
+
+RSpec.describe Event, type: :model do
+  context 'validations' do
+    it { should_not allow_value(nil).for(:name) }
```
santiagovidal: `expect` syntax is preferred

app/models/concerns/event_associator.rb
hdamico: Ok!

PR #32, opened 2020-02-03
Implement ActiveAdmin

### LINKED ISSUE #31: Implement ActiveAdmin
## Description

Implement ActiveAdmin, we should be able to log in as an Admin.

hdamico:
## What does this PR do?

Implements ActiveAdmi, fixes some specs and removes `data` column from `ReviewRequest`

Resolves #31


spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+#
+#  index_admin_users_on_email                 (email) UNIQUE
+#  index_admin_users_on_reset_password_token  (reset_password_token) UNIQUE
+#
+
+require 'rails_helper'
+
+RSpec.describe AdminUser, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
horacio: Let's remove this file from the PR _if_ we won't add specs for this model at this time.

config/initializers/active_admin.rb
```diff
@@ -0,0 +1,328 @@
+ActiveAdmin.setup do |config|
+  # == Site Title
+  #
+  # Set the title that is displayed on the main layout
+  # for each of the active admin pages.
+  #
+  config.site_title = 'Github Analyzer'
```
santiagovidal: Can we rename to: `Code Review Metrics`?

PR #41, opened 2020-02-05
Implement Project model and huge changes!

### LINKED ISSUE #40: Add Project model
## Description

Add `Project` model with `Name`, `Lang` and `Description`

## Things to do

* Generate the model
* Generate Specs
* Relate it with `Events`

hdamico:
## What does this PR do?

Creates `Project` model, and changes how the data flow between the models.

Resolves #40


app/models/events/pull_request.rb
```diff
@@ -40,44 +40,96 @@ class PullRequest < ApplicationRecord
...
-      end
+    def open
+      opened
+    end
 
-      def closed(payload)
-        PullRequestJobs::Closed.perform_later(payload)
-      end
+    def closed
+      close
```
santiagovidal: Why have this method instead of doing everything inside of the `closed` method? Same comment applies for the other actions.

spec/models/event_spec.rb
```diff
@@ -34,15 +36,21 @@
...
-        described_class.resolve(event: '')
-      }.to have_enqueued_job(EventJob)
+        subject.resolve
+      }.to change(Event, :count).by(1)
     end
 
     it 'calls event class' do
-      expect(Events::PullRequest).to receive(:resolve).with(event: 'pull_request')
-      described_class.resolve(event: 'pull_request')
+      expect(subject).to receive(:resolve)
```
santiagovidal: I don't understand this test. Isn't it trivial?

You expect `subject` to receive `resolve` and then call `subject.resolve`.

spec/models/events/pull_request_spec.rb
```diff
@@ -76,45 +96,70 @@
...
       end
     end
 
-    describe '#merged' do
-      it 'enqueues merged' do
+    describe '#review_request_removed' do
+      let!(:pull_request) { create :pull_request_with_review_requests, payload: raw_payload }
+
+      it 'sets status to removed' do
+        review_request = User.find_by!(github_id: raw_payload[:requested_reviewer][:id])
```
santiagovidal: Can't you get the `review_request` from the `pull_request`? Isn't that easier?

spec/models/project_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    it { is_expected.to validate_uniqueness_of(:github_id) }
+    it { is_expected.to have_many(:events) }
+  end
+
+  context '#resolve' do
+    let(:payload) { { repo: { name: 'rs-code-review-metrics', id: 1 } } }
+
+    it 'creates a project' do
+      allow_any_instance_of(Event).to receive(:resolve)
```
santiagovidal: Why do we have this?

app/models/events/pull_request.rb
hdamico: Just to keep a record of the methods on the public side, but it is ok to delete them.

spec/models/event_spec.rb
hdamico: Yes! My bad

spec/models/events/pull_request_spec.rb
hdamico: At this point `pull_request` has 5 `review_requests` and i have to find the `review_request` which has the `reviewer` with `github_id` `raw_payload[:requested_reviewer][:id]`

spec/models/project_spec.rb
hdamico: This is to avoid having errors in the `Event.new` from the `Project` model, we don't want to test that, what we want to test is if it creates a new `Project`. So an easy way is using Mock tests.

app/models/events/pull_request.rb
```diff
@@ -40,44 +40,76 @@ class PullRequest < ApplicationRecord
...
+
+    def open
+      open!
+      save!(opened_at: Time.current)
+    end
+
+    def merged
+      merged!
+      save!(merged_at: Time.current)
+    end
```
lucasocon: Is this for handling `merged` actions? Iirc there is no merged action, but rather a closed action with a merged param = true 
https://developer.github.com/v3/activity/events/types/#pullrequestevent

lucasocon: Be sure to check whether or not merging changes the closed_at attr as well

hdamico: Yes! You are right, thanks

Gemfile
```diff
@@ -40,11 +40,14 @@ gem 'sidekiq', '~> 6.0.2'
 gem 'activeadmin', '~> 2.6.0'
 gem 'devise', '~> 4.7.1'
 
+gem 'dotenv-rails'
```
santiagovidal: Always specify versions in the Gemfile 🙂

app/models/events/pull_request.rb
```diff
@@ -55,9 +55,14 @@ def resolve
     private
 
     def handle_action
+      check_merged
```
lucasocon: What do you think about calling this when the action is 'closed' only?

PR #42, opened 2020-02-06
Test

hdamico:
## What does this PR do?

This PR is only to test webhooks.


horacio:
@hdamico Can we close this?

hdamico:
> @hdamico Can we close this?

We currently use it to test the webhooks, for example, by adding a reviewer, commenting, adding a review, etc.
But we should be able to test it with Rspec, so I will close it.

PR #46, opened 2020-02-07
Review comment event handled

### LINKED ISSUE #44: Implement Review Comment event
## Description

Handle `ReviewComment` event

## Things to do

* Generate Review model
* Associate with Event
* Specs
* Handle different actions

hdamico:
## What does this PR do?

Enables to handle `ReviewComment`'s

Resolves #44


app/models/events/review_comment.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    def find_or_create_review_comment(payload)
+      rc_data = payload['comment']
+      rc = Events::ReviewComment.find_or_create_by!(github_id: rc_data['id']) do |rcom|
+        rcom.owner = find_or_create_user(rc_data['user'])
+        rcom.pull_request = Events::PullRequest.find_by!(github_id: payload['pull_request']['id'])
+      end
+      rc.assign_attributes(payload: payload)
+      rc
```
glm4: Please avoid uncommunicative variable names. Maybe `comment`?

app/models/events/review_comment.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    def created
+      update!(body: payload['comment']['body'])
+    end
+
+    def edited
+      update!(body: payload['changes']['body'])
+    end
+
+    def deleted
```
glm4: same here for the name of the methods.

db/migrate/20200206203510_create_review_comments.rb
```diff
@@ -0,0 +1,12 @@
+class CreateReviewComments < ActiveRecord::Migration[6.0]
+  def change
+    create_table :review_comments do |t|
+      t.integer :github_id
+      t.string :body
+
+      t.timestamps
+    end
+
+    add_column :review_comments, :status, :status, default: 'active'
```
glm4: Any specific reason to add the column outside of the table creation block?

hdamico: The reason is that the type 'status' is a custom type, so if you want to put `t.state` will return an exception that `state` is not a valid type.

app/models/events/review_comment.rb
hdamico: The name of the methods is the same as the name of the action that we receive from github, another way to handle this would be by change `ACTIONS` to a hash where it maps the name of the action as the key and our method as the value.

app/models/events/pull_request.rb
```diff
@@ -61,17 +63,22 @@ def opened(payload)
...
+    end
 
-      def review_request_removed(payload)
-        PullRequestJobs::ReviewRequestRemoved.perform_later(payload)
-      end
+    def merged
+      update!(merged_at: Time.current)
+    end
+
+    def closed
```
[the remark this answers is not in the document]
hdamico: The name of the methods is the same as the name of the action that we receive from github, another way to handle this would be by change ACTIONS to a hash where it maps the name of the action as the key and our method as the value.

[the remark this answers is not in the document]
hdamico: Actually this file should be discarded from this pr, this was solved in this [pr](https://github.com/rootstrap/rs-code-review-metrics/pull/41)

db/migrate/20200206203510_create_review_comments.rb
glm4: Just curious, what are you storing there? Why is a custom type needed?

hdamico:
@glm4 is a way to create a custom enum inside postgres, check this [file](https://github.com/rootstrap/rs-code-review-metrics/blob/master/db/migrate/20200122143238_add_status_type_and_column.rb)

glm4:
> @glm4 is a way to create a custom enum inside postgres, check this [file](https://github.com/rootstrap/rs-code-review-metrics/blob/master/db/migrate/20200122143238_add_status_type_and_column.rb)

That's cool! Any out of the box advantage of doing it this way?

hdamico:
> > @glm4 is a way to create a custom enum inside postgres, check this [file](https://github.com/rootstrap/rs-code-review-metrics/blob/master/db/migrate/20200122143238_add_status_type_and_column.rb)
> 
> That's cool! Any out of the box advantage of doing it this way?

The advantage is that we add a validation on the database side.

PR #49, opened 2020-02-10
Heroku setup and some fixes

### LINKED ISSUE #48: Setup for Heroku
## Description

Set everything up to run properly on Heroku.

## Things to do

* Setup redis
* Add dyno for Sidekiq

hdamico:
## What does this PR do?

Setups everything to run properly on Heroku and fixes some tests and code (just a little)

Resolves #48

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,5 +4,3 @@ GITHUB_ANALYZER_PASSWORD=postgres
 
 # Must be set as the webhook secret on your github app
 GITHUB_ANALYZER_WEBHOOK_SECRET=fake-secret
-REDIS_URL=heroku_redis_url
-SIDEKIQ_THREADS=3
--- a/.env.example
+++ b/.env.example
@@ -4,3 +4,5 @@ GITHUB_ANALYZER_PASSWORD=postgres
 
 # Must be set as the webhook secret on your github app
 GITHUB_ANALYZER_WEBHOOK_SECRET=fake-secret
+REDIS_URL=heroku_redis_url
+SIDEKIQ_THREADS=3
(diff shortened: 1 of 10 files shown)


config/application.rb
```diff
@@ -20,5 +20,7 @@ class Application < Rails::Application
     config.active_record.schema_format = :sql
     # Autoload lib folder
     config.autoload_paths << "#{Rails.root}/lib"
+
+    config.hosts << 'code-review-metrics.herokuapp.com'
```
santiagovidal: This should be an Env variable.

Procfile
```diff
@@ -0,0 +1,2 @@
+web: bin/rails server -p $PORT -e $RAILS_ENV
```
juanmanuelramallo: This should actually use puma

```suggestion
web: bundle exec puma -C config/puma.rb
```

https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#procfile

.env.example
```diff
@@ -4,3 +4,5 @@ GITHUB_ANALYZER_PASSWORD=postgres
 
 # Must be set as the webhook secret on your github app
 GITHUB_ANALYZER_WEBHOOK_SECRET=fake-secret
+REDIS_URL=heroku_redis_url
+SIDEKIQ_THREADS=3
```
horacio: Aren't we setting the concurrency twice? Here and then in the Procfile? 👀

hdamico: Not really, the worker dyno actually is not working, i should delete that dyno.

PR #52, opened 2020-02-13
Implement review event

### LINKED ISSUE #43: Add Review event
## Description

Handle Review event

## Things to do

* Generate Review model
* Associate with Event
* Specs
* Handle different actions

hdamico:
## What does this PR do?

Adds `Review` model with their actions

Resolves #43

### DIFF
--- a/app/models/events/review.rb
+++ b/app/models/events/review.rb
@@ -1,6 +1,6 @@
 # == Schema Information
 #
-# Table name: review_comments
+# Table name: reviews
 #
 #  id              :bigint           not null, primary key
 #  body            :string
@@ -13,8 +13,8 @@
 #
 # Indexes
 #
-#  index_review_comments_on_owner_id         (owner_id)
-#  index_review_comments_on_pull_request_id  (pull_request_id)
+#  index_reviews_on_owner_id         (owner_id)
+#  index_reviews_on_pull_request_id  (pull_request_id)
 #
 # Foreign Keys
 #
@@ -23,17 +23,17 @@
 #
 
 module Events
-  class ReviewComment < ApplicationRecord
-    ACTIONS = %w[created edited deleted].freeze
+  class Review < ApplicationRecord
+    ACTIONS = %w[submitted edited dismissed].freeze
 
     enum status: { active: 'active', removed: 'removed' }
 
     has_many :events, as: :handleable, dependent: :destroy
     belongs_to :owner, class_name: 'User',
                        foreign_key: :owner_id,
-                       inverse_of: :owned_review_comments
+                       inverse_of: :owned_reviews
     belongs_to :pull_request, class_name: 'Events::PullRequest',
-                              inverse_of: :review_requests
+                              inverse_of: :reviews
 
     validates :status, inclusion: { in: statuses.keys }
     validates :github_id, presence: true
@@ -59,14 +59,14 @@ module Events
       end
     end
 
-    def find_or_create_review_comment(payload)
-      comment_data = payload['comment']
-      comment = Events::ReviewComment.find_or_create_by!(github_id: comment_data['id']) do |rc|
-        rc.owner = find_or_create_user(comment_data['user'])
+    def find_or_create_review(payload)
+      review_data = payload['review']
+      review = Events::Review.find_or_create_by!(github_id: review_data['id']) do |rc|
+        rc.owner = find_or_create_user(review_data['user'])
         rc.pull_request = Events::PullRequest.find_by!(github_id: payload['pull_request']['id'])
       end
-      comment.assign_attributes(payload: payload)
-      comment
+      review.assign_attributes(payload: payload)
+      review
     end
 
     def handleable?
@@ -75,15 +75,15 @@ module Events
 
     # Actions
 
-    def created
-      update!(body: payload['comment']['body'])
+    def submitted
+      update!(body: payload['review']['body'])
     end
 
     def edited
       update!(body: payload['changes']['body'])
     end
 
-    def deleted
+    def dismissed
       removed!
     end
   end
--- a/db/migrate/20200212151614_create_reviews.rb
+++ b/db/migrate/20200212151614_create_reviews.rb
@@ -1,0 +1,13 @@
+class CreateReviews < ActiveRecord::Migration[6.0]
+  def change
+    create_table :reviews do |t|
+      t.references :pull_request, null: false, foreign_key: true
+      t.references :owner, foreign_key: { to_table: :users }
+      t.integer :github_id
+      t.string :body
+
+      t.timestamps
+    end
+    add_column :reviews, :status, :status, default: 'active'
+  end
+end
--- a/spec/models/events/review_comment_spec.rb
+++ b/spec/models/events/review_comment_spec.rb
@@ -84,7 +84,7 @@ RSpec.describe Events::ReviewComment, type: :model do
     end
 
     describe '#created' do
-      before { subject.payload['action'] = 'created' }
+      before { change_action_to 'created' }
 
       it 'sets body' do
         expect {
@@ -94,7 +94,7 @@ RSpec.describe Events::ReviewComment, type: :model do
     end
 
     describe '#edited' do
-      before { subject.payload['action'] = 'edited' }
+      before { change_action_to 'edited' }
 
       it 'edits body' do
         body = payload['comment']['body']
@@ -106,7 +106,7 @@ RSpec.describe Events::ReviewComment, type: :model do
     end
 
     describe '#deleted' do
-      before { subject.payload['action'] = 'deleted' }
+      before { change_action_to 'deleted' }
 
       it 'sets removed' do
         expect {
--- a/spec/models/events/review_comment_spec.rb
+++ b/spec/models/events/review_comment_spec.rb
@@ -97,11 +97,11 @@ RSpec.describe Events::ReviewComment, type: :model do
       before { change_action_to 'edited' }
 
       it 'edits body' do
-        body = payload['comment']['body']
-        subject.update!(body: body)
+        comment_body = payload['comment']['body']
+        subject.update!(body: comment_body)
         expect {
           subject.resolve
-        }.to change { subject.reload.body }.from(body).to(payload['changes']['body'])
+        }.to change { subject.reload.body }.from(comment_body).to(payload['changes']['body'])
       end
     end
 
--- a/spec/models/events/review_comment_spec.rb
+++ b/spec/models/events/review_comment_spec.rb
@@ -37,6 +37,11 @@ RSpec.describe Events::ReviewComment, type: :model do
       expect(subject).to_not be_valid
     end
 
+    it 'is not valid without body' do
+      subject.body = nil
+      expect(subject).to_not be_valid
+    end
+
     it { is_expected.to belong_to(:pull_request) }
   end
 
@@ -80,16 +85,6 @@ RSpec.describe Events::ReviewComment, type: :model do
         subject.save!
 
         expect(subject.send(:find_or_create_review_comment, payload)).to eq(subject)
-      end
-    end
-
-    describe '#created' do
-      before { change_action_to 'created' }
-
-      it 'sets body' do
-        expect {
-          subject.resolve
-        }.to change { subject.reload.body }.from(nil).to(payload['comment']['body'])
       end
     end
 
--- a/spec/models/events/review_spec.rb
+++ b/spec/models/events/review_spec.rb
@@ -84,7 +84,7 @@ RSpec.describe Events::Review, type: :model do
     end
 
     describe '#submitted' do
-      before { subject.payload['action'] = 'submitted' }
+      before { change_action_to 'submitted' }
 
       it 'sets body' do
         expect {
@@ -94,7 +94,7 @@ RSpec.describe Events::Review, type: :model do
     end
 
     describe '#edited' do
-      before { subject.payload['action'] = 'edited' }
+      before { change_action_to 'edited' }
 
       it 'edits body' do
         body = payload['review']['body']
@@ -106,7 +106,7 @@ RSpec.describe Events::Review, type: :model do
     end
 
     describe '#dismissed' do
-      before { subject.payload['action'] = 'dismissed' }
+      before { change_action_to 'dismissed' }
 
       it 'sets removed' do
         expect {
--- a/spec/models/events/review_spec.rb
+++ b/spec/models/events/review_spec.rb
@@ -1,6 +1,6 @@
 # == Schema Information
 #
-# Table name: review_comments
+# Table name: reviews
 #
 #  id              :bigint           not null, primary key
 #  body            :string
@@ -13,8 +13,8 @@
 #
 # Indexes
 #
-#  index_review_comments_on_owner_id         (owner_id)
-#  index_review_comments_on_pull_request_id  (pull_request_id)
+#  index_reviews_on_owner_id         (owner_id)
+#  index_reviews_on_pull_request_id  (pull_request_id)
 #
 # Foreign Keys
 #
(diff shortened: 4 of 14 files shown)


spec/models/events/review_comment_spec.rb
```diff
@@ -26,6 +26,17 @@
 
 RSpec.describe Events::ReviewComment, type: :model do
   describe 'validations' do
+    subject { build :review_comment }
+
+    it 'is valid with valid attributes' do
+      expect(subject).to be_valid
+    end
+
+    it 'is not valid without github id' do
```
mrubi-rootstrap: This is a question. Are there any other significant fields values that might invalidate a ReviewComment, like an empty comment text?

spec/models/events/review_comment_spec.rb
```diff
@@ -73,27 +84,33 @@
     end
 
     describe '#created' do
+      before { subject.payload['action'] = 'created' }
+
       it 'sets body' do
         expect {
-          subject.send(:created)
+          subject.resolve
         }.to change { subject.reload.body }.from(nil).to(payload['comment']['body'])
```
mrubi-rootstrap: @hdamico Would it make sense to define the following expressions

```
      let(:payload_comment) { payload.comment }
      let(:payload_comment_body) { payload_comment.body }
```

and change the occurrences of

`payload['comment']['body']`

to

`payload_comment_body` ?

I think it might improve a bit the test ligibility.

spec/models/events/review_comment_spec.rb
```diff
@@ -26,6 +26,17 @@
...
+    it 'is valid with valid attributes' do
+      expect(subject).to be_valid
+    end
+
+    it 'is not valid without github id' do
+      subject.github_id = nil
+      expect(subject).to_not be_valid
+    end
+
     it { is_expected.to belong_to(:pull_request) }
```
mrubi-rootstrap: This is a question.

This test skips its description. Is it a style convention or would

```
it 'belongs to a PullRequest' do
   is_expected.to belong_to(:pull_request)
end
```

be more declarative?

spec/models/events/review_comment_spec.rb
```diff
@@ -73,27 +84,33 @@
...
+          subject.resolve
         }.to change { subject.reload.body }.from(nil).to(payload['comment']['body'])
       end
     end
 
     describe '#edited' do
+      before { subject.payload['action'] = 'edited' }
+
       it 'edits body' do
         body = payload['comment']['body']
```
mrubi-rootstrap: @hdamico, `payload` has several nested fields named `body`.
Please rename the variable `body` to `comment_body` to disambiguate its intention.

spec/models/events/review_comment_spec.rb
```diff
@@ -73,27 +84,33 @@
...
       it 'sets body' do
         expect {
-          subject.send(:created)
+          subject.resolve
         }.to change { subject.reload.body }.from(nil).to(payload['comment']['body'])
       end
     end
 
     describe '#edited' do
+      before { subject.payload['action'] = 'edited' }
```
mrubi-rootstrap: This is a style question.
Would it make sense to define a method

```
def set_payload_action_to(action)
   subject.payload['action'] = action
end
```

at the beginning of the spec file and to change

```
before { subject.payload['action'] = 'edited' }
```

to

```
before { set_payload_action_to 'edited' }
```

?

spec/models/events/review_comment_spec.rb
```diff
@@ -73,27 +84,33 @@
...
         subject.update!(body: body)
         expect {
-          subject.send(:edited)
+          subject.resolve
         }.to change { subject.reload.body }.from(body).to(payload['changes']['body'])
       end
     end
 
     describe '#deleted' do
+      before { subject.payload['action'] = 'deleted' }
```
mrubi-rootstrap: Same question than for the 'edited' case.

spec/models/events/review_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      }.deep_stringify_keys
+    end
+
+    describe '#find_or_create_review' do
+      subject { build :review }
+      before { create :pull_request, github_id: payload['pull_request']['id'] }
+
+      it 'creates a review' do
+        expect {
+          subject.send(:find_or_create_review, payload)
```
mrubi-rootstrap: This is a question. Would it be possible to call the method

`subject.find_or_create_review(payload)`

instead or is it a private method?

spec/models/events/review_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      before { subject.payload['action'] = 'dismissed' }
+
+      it 'sets removed' do
+        expect {
+          subject.resolve
+        }.to change { subject.reload.status }.from('active').to('removed')
+      end
+    end
+  end
+end
```
mrubi-rootstrap: @hdamico I did not repeat the same questions than for `reviews_spec` but they hold for this spec as well.

app/models/events/review.rb
```diff
@@ -0,0 +1,90 @@
...
+      review
+    end
+
+    def handleable?
+      ACTIONS.include?(payload['action'])
+    end
+
+    # Actions
+
+    def submitted
```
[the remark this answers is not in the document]
hdamico: Whenever you create a method with the '!' mark means that there is another method without the '!', so in this case we just have one.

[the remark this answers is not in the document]
hdamico: Same as i commented before. Anyway check that methods should be with `underscore` and not camelcase

spec/models/events/review_comment_spec.rb
hdamico: I didn't understand your point, do you mean the variable name?

spec/models/events/review_comment_spec.rb
hdamico: Yes! That would be great. Thanks!

spec/models/events/review_spec.rb
hdamico: Actually is private method, i understand that calling private methods in specs goes against TDD, but i didn't want to have to call the Event model to check that. If you have or know any other way to test that, it would be appreciated.

spec/models/events/review_spec.rb
hdamico: What do you mean?

app/models/events/review.rb
```diff
@@ -0,0 +1,90 @@
...
+#
+#  fk_rails_...  (owner_id => users.id)
+#  fk_rails_...  (pull_request_id => pull_requests.id)
+#
+
+module Events
+  class Review < ApplicationRecord
+    ACTIONS = %w[submitted edited dismissed].freeze
+
+    enum status: { active: 'active', removed: 'removed' }
```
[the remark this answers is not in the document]
hdamico: Mmm... We have those 3 actions, in none of them tells you if the review was approved.

spec/models/events/review_comment_spec.rb
mrubi-rootstrap: Yes, I meant the variable name, to make it explicit that is the body of the comment.

app/models/events/review.rb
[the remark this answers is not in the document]
mrubi-rootstrap: All right. Thanks!
Closing this comment.

spec/models/events/review_spec.rb
mrubi-rootstrap: How about something like

```
  before do
    def subject.find_or_create_review(payload)
      super(payload)
    end
  end
```

?

I just came up with it, I am not sure whether it could be convenient or not.

mrubi-rootstrap: Maybe an Rspec helper could make it explicit that the test intention is to send a private message to an object

```
before do
   enable_private_method :find_or_create_review, on: subject
end
```

app/models/events/review.rb
```diff
@@ -0,0 +1,90 @@
...
+    def resolve
+      return unless handleable?
+
+      handle_action
+    end
+
+    private
+
+    def handle_action
+      send(payload['action'])
```
mrubi-rootstrap: Is it possible to change the use of `send` by an explicit switch or hash?

```
event_action = payload['action']

if event_action == 'submitted'
   submitted
elif event_action == 'edited'
   edited
elif event_action == 'dismissed'
   dismissed
end
```

or may be

```
event_action = payload['action']

handled_action = Hash[
   'submitted' => proc{ submitted },
   'edited' => proc{ edited },
   'dismissed' => proc{ dismissed },
]

handled_action[event_action].call
```

hdamico: Is not really perfomant doing that. I am using the `send` inside a private method, so in that case it should not be a problem. Why would you prefer doing that?

spec/models/events/review_comment_spec.rb
hdamico: Mmm.. In that context i don't think is it necessary to put `comment_body`. We know that we are in the `ReviewComment` model, and in the describe says that it changes the body.

mrubi-rootstrap: I was thinking in reading

`subject.update!(body: body)`

or

`}.to change { subject.reload.body }.from(body).to(payload['changes']['body'])`

in isolation.

While it is true that the context disambiguates the meaning of the variable, naming it

`subject.update!(body: comment_body)`

or

`}.to change { subject.reload.body }.from(comment_body).to(payload['changes']['body'])`

makes it explicit.

It was a suggestion anyway.

spec/models/events/review_comment_spec.rb
hdamico: Yes! I should add that validation! Thanks!

app/models/events/review.rb
mrubi-rootstrap: I would think that the first alternative using `ifs` is more performant than using `send`. I think the use of `send` requires at least a message send for the `:send` message and another one for the actual message.

The second option is not performant, that is true.

The reasons would be
- to not bind the naming convention of the web hook to a method name. The method name would remain the same even if the action name changes.
- to make an explicit message send for clarity. The use of `:send` makes it harder to find references for message sends.
- to make it explicit that `handle_action` maps some actions to handler methods and to make it explicit how each action is mapped to each handler method

hdamico: The `send` is the same as using `Model.method`. Asking `if` would replace the `handleable` method, which asks if the action that is comming is included in a hash that contains the event that we are able to handle, such as `pull_request`. About the name of the method, we could use aliases if you prefer. And about the understand or explicity of the code, i beleave here we are using a sort of duck typing, we is a nice feature from rails and every dynamic programming language.

spec/models/events/review_comment_spec.rb
hdamico: Ok, i will!

spec/models/events/review_spec.rb
hdamico: I will check if we are already handling this method in the specs without this one. Otherwise i guess we will should make a mock test that starts from the `Event` model where this method is triggered.

db/migrate/20200212151614_create_reviews.rb
```diff
@@ -0,0 +1,13 @@
+class CreateReviews < ActiveRecord::Migration[6.0]
+  def change
+    create_table :reviews do |t|
+      t.references :pull_request, null: false, foreign_key: true
+      t.references :owner, foreign_key: { to_table: :users }
+      t.integer :github_id
+      t.string :body
```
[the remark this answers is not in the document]
hdamico: Do you mean for the body?

santiagovidal:
Shouldn't we track if the review was approved, rejected or just commented?

app/models/events/review.rb
mrubi-rootstrap: It surprises me that the use of `:send` is as efficient as a regular message send but that is an awesome optimisation.
It also surprises me that a conditional `if` is less efficient than testing an element for inclusion in a Set. I would have thought that conditionals are highly optimised whilst testing an element inclusion in a Set requires to calculate the hash value of the element and then perform an `if` to check if the hash is present in the Set.

But I think you have a point.

However I still find the line `send(payload['action'])` a bit cryptic.

How about changing a bit the class protocol for the following:

```
    def resolve
      action = payload['action']

      return unless handleable?(action)

      handle_action(action)
    end

    private

    def handle_action(actionMethod)
      send(actionMethod)
    end

    def handleable?(actionMethod)
      ACTIONS.include?(actionMethod)
    end
```

The rationale for this change is that passing the action as a parameter gives the reader of the code full context in the method `resolve` (the reader would read the question `is the action handleable?` instead of 'is handleable?':

`handleable?(action)` vs `handleable?`

and also helps to understand at a glance that the github action name is being send as a Ruby method in `send(actionMethod)`.

As a side effect it avoids the duplication of the statement `payload['action']`.

hdamico: Whenever you are not sure about a comment or doubting, you can check it by yourself by doing a script with `Benchmark`.
And yes! I am making a refactor where a service will be created to manage the payload and the GitHub workflow. This will make everything more clear.
But now I have to get this in master to make the refactor.

spec/models/events/review_spec.rb
mrubi-rootstrap: @hdamico I took a deeper look at the context of this method call and I think there is change that might improve the separation of concerns and make the code a bit more clear.

If you look at the method

```
def find_or_create_review(payload)
    review_data = payload['review']
    review = Events::Review.find_or_create_by!(github_id: review_data['id']) do |rc|
    rc.owner = find_or_create_user(review_data['user'])
    rc.pull_request = Events::PullRequest.find_by!(github_id: payload['pull_request']['id'])
  end
  review.assign_attributes(payload: payload)
  review
end
```

you can see that even it is an instance method it does not access any instance side data from the receiver. That might point out that the method could be moved to a different place.

Also note that this method creates an instance of the same class than itself, which can be a bit confusing:

  `Review.new.new_from_payload` instead of `Review.new_from_payload`

So the method implementation suggests that it could be moved to the class side.

That would make it a bit clearer since classes responsibility is to create instances and would simplify the code.

In the Event class the method `Event.find_or_create_event` can be written as

```
def find_or_create_event
  const_event.new_from_payload(data)
end
```

and the method `Review.find_or_create_review` can be renamed and moved to the class side with

```
def self.new_from_payload(payload)
  comment_data = payload['comment']
  comment = self.find_or_create_by!(github_id: comment_data['id']) do |rc|
    rc.owner = User.new_from_payload(comment_data['user'])
    rc.pull_request = Events::PullRequest.find_by!(github_id: payload['pull_request']['id'])
  end
  comment.assign_attributes(payload: payload)
  comment
end
```

A side effect of this refactor is that it is no longer necessary to use `:send` and to dynamically build the method name in

```
def find_or_create_event
  const_event.new.send("find_or_create_#{name}", data)
end
```

since the message is replaced by the polymorphic method `new_from_payload` that each event type would implement.

There are several more small refactors that can be done to the method:

```
def self.new_from_payload(payload)
  # All these getters suggest that a service object to abstract the internal structure of
  # the payload might be a great improvement as you suggested in another comment.
  comment_data = payload['comment']
  comment_id = comment_data['id']
  user_data = comment_data['user']
  pull_request_id = payload['pull_request']['id']

  self.find_or_create_by!(github_id: comment_id) do |comment|
    comment.owner = User.new_from_payload(user_data)
    comment.pull_request = Events::PullRequest.find_by!(pull_request_id)
    comment.assign_attributes(payload: payload)
  end
end
```

but I think that for now the most significant change would be to move the creation of the Review instance to the class side.

mrubi-rootstrap: I know this kind of refactors put quite a delay on the merge of the original PR. Sorry about that but I think this refactor is significant.

hdamico: Great! That is a nice approach. Anyway I would change that in other pull request, actually i am making a refactor, and this will be useful.
Thanks!

Issue #53, opened 2020-02-17
Create Metrics Model


mrubi-rootstrap:
Create the model Metrics and its table in the database.

The first schema will have the form

metrics(entity_key, metrics_key, value, value_timestamp).

Example of a metrics record:

entity_key, metric_key, value, timestamp

generis, review_turnaround, 495, null
mmansilla, review_turnaround, 433, null
generis_week, review_turnaround, 223, 03/02/2020
generis_week, review_turnaround, 223, 10/02/2020
generis_month, review_turnaround, 223, 01/02/2020

PR #54, opened 2020-02-17
53 create metrics model

mrubi-rootstrap:
## What does this PR do?

Implements the task [#53 Create Metrics Model](https://github.com/rootstrap/rs-code-review-metrics/issues/53)

It adds the migration to create the table metrics, the metrics Model with the basic validations and its tests.

This PR does not implement the relation between the Metrics and other entities like Users, Comments and PullRequests, just the metrics table.

The relation between the Metrics table and the other entities and the population of the metrics table is to be implemented in a different task.

### DIFF
--- a/db/migrate/20200217165218_create_metrics.rb
+++ b/db/migrate/20200217165218_create_metrics.rb
@@ -1,0 +1,12 @@
+class CreateMetrics < ActiveRecord::Migration[6.0]
+  def change
+    create_table :metrics do |t|
+      t.string :entity_key,         null: false
+      t.string :metric_key,         null: false
+      t.decimal :value,             null: true
+      t.datetime :value_timestamp,  null: true
+
+      t.timestamps
+    end
+  end
+end
--- a/spec/factories/metrics.rb
+++ b/spec/factories/metrics.rb
@@ -1,0 +1,8 @@
+FactoryBot.define do
+  factory :metric do
+    entity_key { 'generis' }
+    metric_key { 'review_turnaround' }
+    value { 495 }
+    value_timestamp { Time.utc(2020,2,1,0,0,0) }
+  end
+end
--- a/spec/factories/metrics.rb
+++ b/spec/factories/metrics.rb
@@ -1,3 +1,16 @@
+# == Schema Information
+#
+# Table name: metrics
+#
+#  id              :bigint           not null, primary key
+#  entity_key      :string           not null
+#  metric_key      :string           not null
+#  value           :decimal(, )
+#  value_timestamp :datetime
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#
+
 FactoryBot.define do
   factory :metric do
     entity_key { 'generis' }
--- a/spec/factories/metrics.rb
+++ b/spec/factories/metrics.rb
@@ -3,6 +3,6 @@ FactoryBot.define do
     entity_key { 'generis' }
     metric_key { 'review_turnaround' }
     value { 495 }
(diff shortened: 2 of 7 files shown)


spec/models/metric_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+        .with_options(null: true)
+    end
+  end
+
+  describe 'validation' do
+    describe 'of entity_key' do
+      it 'fails with a missing value' do
+        subject.entity_key = nil
+
+        expect(subject).to_not be_valid
```
glm4: doesn't `validate_presence_of` work for you?

spec/models/metric_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      it 'fails with a value too long' do
+        subject.entity_key = 'a' * 256
+
+        expect(subject).to_not be_valid
```
glm4: same but with `validate_length_of`

spec/models/metric_spec.rb
mrubi-rootstrap: Did not know about it. Will try it tomorrow. Thanks!

spec/factories/metrics.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :metric do
+    entity_key { 'generis' }
```
santib: I usually have fake data in my factories so my tests don't assume values from the factories (so they don't get coupled). What do you think?

db/migrate/20200217165218_create_metrics.rb
```diff
@@ -0,0 +1,12 @@
+class CreateMetrics < ActiveRecord::Migration[6.0]
+  def change
+    create_table :metrics do |t|
+      t.string :entity_key,         null: false
```
santib: should we add indexes for `entity_key`, `metric_key` and `value_timestamp`?

spec/factories/metrics.rb
hdamico: Yes, a good option is `Faker`. We should add it in all the factories, but that could be in another PR.

spec/models/metric_spec.rb
mrubi-rootstrap: Done. Thanks!

spec/models/metric_spec.rb
mrubi-rootstrap: Done. Thanks!

spec/factories/metrics.rb
mrubi-rootstrap: @santib I agree that if a test asserts on a particular value it should be override the value in the same test instead of relying on the factory value.

@hdamico  Should I create a new ticket in the board for that?

db/migrate/20200217165218_create_metrics.rb
mrubi-rootstrap: I am not sure about that yet.

I don't have a clear idea yet of how is this table to be accessed in the actual use cases. It is possible that it might be joined by its primary key alone.

As this table is meant to have a lot of records and each index might add a significant space overhead to the table structure I would like to postpone the creation of indices until the use cases are implemented. Would that be possible?

santib: Sure

spec/factories/metrics.rb
hdamico: That would be appreciated!

mrubi-rootstrap: Added the task https://github.com/rootstrap/rs-code-review-metrics/issues/56 to the backlog.

Would you please review it when you have some time?

Thanks!

Issue #58, opened 2020-02-19
Create a spec/factory to create GitHub event payloads


mrubi-rootstrap:
# Description

Create one or more `spec/factory` definitions with common payloads used in the Github Webhooks.

# Background

The application listens to Github Webhooks.

Each Webhook passes a payload argument with the event data.

The payload argument is a json object.

A description and examples of each payload passed by each Github Webhook can be found [here](https://developer.github.com/v3/activity/events/types/).

As you can see from the [PullRequestReview example](https://developer.github.com/v3/activity/events/types/#pullrequestreviewevent) the payload has a lot of details for each event but not all that information is used by the application.

# Implementation

Define a factory for each Github Webhook payload. For simplicity include in the payload only the data required by the application.

The implementation of the factory has the following catch: by default FactoryBot maps the created object to a class instance named like the factory. For example to define

```
factory :pull_request_review_payload do
  #...
end
```

it looks for a class named PullRequestReviewPayload.

In this case the Factory would create a Hash with the payload information as it would come from the Webhook therefore in the factory definition set the class to `Hash`:

```
FactoryBot.define do
  factory :pull_request_review_payload, class: Hash do
    action. { 'submitted' }
    id. { 237895671 }
    node_id. { 'MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjM3ODk1Njcx' }
    state. { 'commented' }

    initialize_with do
      {
          'action' => action,
          'review' => {
            'id' =>  id,
            'node_id' =>  node_id
          },
          'state' => state
      }
    end
  end
end
```

Another suggestion for the implementation is to use FactoryBot [traits](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#traits) for the possible enum values a payload might have, like the event fields `action` and `state`.

mrubi-rootstrap:
Closing this task as it is being address in #60

PR #59, opened 2020-02-19
58 create payload factory

mrubi-rootstrap:
## What does this PR do?

Implement task [58-Create a spec/factory to create GitHub event payloads](https://github.com/rootstrap/rs-code-review-metrics/issues/58)

### Details

This PR creates a factory to generate the Github Webhooks payloads templates for the tests.

These payloads are json objects and the tests of the process that generates metrics will likely create and use several these payload templates during its tests.


mrubi-rootstrap:
Closing this PR without merging as it's being addressed in #60

PR #60, opened 2020-02-19
Refactor

### LINKED ISSUE #66: Code Refactor
## Description

Move GitHub logic from the models to services.

hdamico:
## What does this PR do?

Gets GitHub logic out of the models to services.

Resolves #66


hdamico:
### It is still in progress
## Things to do:
- [ ] Add specs for `EventBuilders`
- [ ] Add Faker gem to improve Factories

Any suggest please let me know.

spec/services/github_service.rb
```diff
@@ -0,0 +1,130 @@
+require 'rails_helper'
+
+RSpec.describe GithubService do
+  subject { described_class.call(payload: payload, event: event) }
+
+  describe 'events' do
+    context 'pull request' do
+      let!(:payload) { (create :pull_request_payload).merge((create :repository_payload)) }
```
mrubi-rootstrap: `create :object` is used by FactoryBot to build a model and insert it in the database.

To build a model without inserting it in the database it uses `build :object`.

For payloads it would be the same because they are not ActiveRecords but to preserve the semantics of FactoryBot I would change `create` for `build`.

What do you think?

mrubi-rootstrap: This line

`let!(:payload) { (create :pull_request_payload).merge((create :repository_payload)) }`

got me confused.

The reason is that when reading it I thought there was a logical error in merging the attributes of a pull_request with the attributes of a repository.

Going back to the factory source code I saw that the merge was ok because `create :pull_request_payload` returns 

```
pull_request: {
  id: ...
  ...
}
repository: {
  id: ...
}
```

That made me think that the factory should not include the first payload key but only the attributes:

`create :pull_request_payload`

`{
  id: ...
}`

The reason is that it would allow to insert a payload object into another payload object using a different name.

Here is an example:

If `build :user_payload` returns `{ id: ... }` instead of `{ user: { id... }` it allows to use like this:

```
pull_request_event = {
   action: ...,
   pull_request: {
      user: (build :user, id: 1)
   },
   requested_reviewer: (build :user, id: 2)
}
```

Including only the attributes of a payload object would also be consistent with the behaviour of FactoryBot when it builds a model object.

app/models/events/pull_request.rb
```diff
@@ -23,15 +23,13 @@
 #
 module Events
   class PullRequest < ApplicationRecord
-    ACTIONS = %w[open review_requested closed \
-                 review_request_removed].freeze
-    private_constant :ACTIONS
-
     enum state: { open: 'open', closed: 'closed' }
```
mrubi-rootstrap: This is a style question that applies to all model definitions as well.

Would it be possible to add the declaration of each attribute to the class definition?

Even though it is not necessary the rationale is the class to self document its attributes and allow a reader of the code to know at a glance which attributes does an ActiveRecord subclass has without looking into the migration, the database schema or the tests.

It would also declare each attribute configuration like being nullable or having a default value.

app/services/action_handler.rb
```diff
@@ -0,0 +1,17 @@
...
+    @event = event
+    @event_type = event_type
+  end
+
+  def call
+    resolve
+  end
+
+  private
+
```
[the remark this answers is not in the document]
hdamico: I don't get why would we need to implement this.
```
def handleable?
  raise "Subclass #{self.class.name} must implement the method :handleable?"
end
```
Rails by himself has his own Exception when a method is not implemented.

And this:
```
def handled_actions
  self.class.ACTIONS
end

def handleable?
  handled_actions.include?(@payload['action'])
end
```
Won't work, because we can't access to constants with `self.class.ACTIONS`.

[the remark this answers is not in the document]
mrubi-rootstrap: Having a different exception from `missing_method` is solely for documentation purposes.

A missing method can be raised for different reasons:

- the sender could be sending the wrong message to the right object
- the sender could be sending the right message to the wrong object
- the sender could be sending the right message to the right object but the error is on the receiver side

A `raise ObjectShouldRespondToMessage.new` documents that a missing method would occur in the context of third scenario. If a developer is creating a new class that should comply with the ActionHandler protocol she would get the error stating that the method should be implemented and will have a bit more context to implement the class.

From the `ActionHandler` protocol side it allows a reader to know which methods does the ActionHandler expects to be implemented in a different place.

Probably a better implementation would be

```
def handleable?
  raise ObjectShouldRespondToMessage.new
end
```

and yet a more sophisticated approach could be something like

```
class ActionHandler < BaseService
  extend Interface

  object_should_respond_to :handleable?
end
```

The reasons are arguable, of course.

app/services/event_builders/pull_request.rb
```diff
@@ -0,0 +1,15 @@
...
+  class PullRequest < EventBuilder
+    ATTR_PAYLOAD_MAP = { github_id: 'id', number: 'number', state: 'state', node_id: 'node_id',
+                         title: 'title', locked: 'locked', draft: 'draft' }.freeze
+    def build
+      pr_data = @payload['pull_request']
+
+      Events::PullRequest.find_or_create_by!(github_id: pr_data['id']) do |pr|
+        pr.attributes = ATTR_PAYLOAD_MAP.inject({}) do |hash, (key, _v)|
+          hash.merge!(key => pr_data.fetch(ATTR_PAYLOAD_MAP.fetch(key)))
+        end
```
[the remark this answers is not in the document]
hdamico: I made my own benchmarks and I got this
```
# Tap
0.004965   0.000426   0.005391 (  0.006454)

# Send
0.004417   0.000445   0.004862 (  0.005944)

# Inject
0.004729   0.000418   0.005147 (  0.006121)
```
The `send` one, which was the best, is doing this
`ATTR_PAYLOAD_MAP.map { |key, value| pr.public_send("#{key}=", pr_data.fetch(value)) }`
So what do you think?

app/services/action_handler.rb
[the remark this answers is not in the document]
mrubi-rootstrap: I missed that ACTIONS would not be accessible, sorry.

Still may be it would be possible to create an accessor method in the subclass:

```
class PullRequest
  # Returns an Array with all the payload actions the receiver handles.
   def handled_actions
      ACTIONS
   end
end
```

and the super class would use it like

```
class ActionHandler < BaseService
  ...
  private

  # Returns an Array with all the payload actions the receiver handles.
  # Each action is expected to be a String.
  def handled_actions
    raise ObjectShouldRespondToMessageError.new
  end

  # Returns true if the receiver handles the action of the payload it holds, false otherwise.
  def handleable?
    handled_actions.include?(@payload['action'])
  end

  # Handles the action of the payload it holds.
  # Assumes the action is :handleable?. This assumption must be validated prior to calling
  # this method.
  def handle_action
    send(@payload['action'])
  end
end
```

app/services/event_builders/pull_request.rb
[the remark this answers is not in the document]
mrubi-rootstrap: Nice! easier to read than the previous ones.

How about

```
ATTR_PAYLOAD_MAP.each_pair { |attribute_key, payload_key|
  pr.write_attribute(attribute_key, pr_data.fetch(payload_key))
}
```

?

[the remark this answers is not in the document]
hdamico:
```
# Write Attributes
0.005076   0.000449   0.005525 (  0.006665)
```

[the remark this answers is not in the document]
mrubi-rootstrap: All right.

If the optimisation is worth it I would go for the most efficient solution and include a comment stating that the use of `map` and `public_send("#{key}=")` are part of an optimization.

app/services/action_handler.rb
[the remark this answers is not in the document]
hdamico: I am trying to understand what you said, what do you want to implement is an `Abstract Factory`?
Because otherwise we are overloading the app with unnecessary code, Ruby code should be self-document. Also, we are skipping Rails exception control.
If you are trying to implement Abstract Factory, Rails already does it in the `Base` class. And if you want to log it, you should use a `Rescue` clause with a custom message.

app/services/action_handlers/pull_request.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def closed
+      merged if @payload['pull_request']['merged'] == true
+      @event_type.closed!
+      @event_type.update!(closed_at: Time.current)
+    end
+
+    def review_request_removed
+      reviewer = User.find_by!(github_id: @payload['requested_reviewer']['id'])
```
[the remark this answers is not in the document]
hdamico: Ok

app/services/action_handlers/pull_request.rb
```diff
@@ -0,0 +1,46 @@
...
+      @event_type.review_requests.find_by!(reviewer: reviewer).removed!
+    end
+
+    def review_requested
+      owner = find_or_create_user(@payload['pull_request']['user'])
+      reviewer = find_or_create_user(@payload['requested_reviewer'])
+      @event_type.review_requests.create!(owner: owner, reviewer: reviewer,
+                                          node_id: reviewer.node_id, login: reviewer.login)
```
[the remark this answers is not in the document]
hdamico: I believe it is the main purpose of the method, to create a `review_request`, and is in the only part of the app where is being used.

app/services/github_service.rb
```diff
@@ -0,0 +1,48 @@
+class GithubService < BaseService
+  def initialize(payload:, event:)
+    @payload = payload
+    @event = event
```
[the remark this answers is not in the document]
hdamico: Remember that in Ruby everything is an `Object`, so initializing a variable with a `NilClass` which maybe is not going to be used, is a waste of memory.

db/migrate/20200212151614_create_reviews.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :reviews do |t|
+      t.references :pull_request, null: false, foreign_key: true
+      t.references :owner, foreign_key: { to_table: :users }
+      t.integer :github_id
+      t.string :body
+
+      t.timestamps
+    end
+    add_column :reviews, :status, :status, default: 'active'
```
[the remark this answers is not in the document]
hdamico: `Status` is a custom type, so you can't use it inside a `create_table` block.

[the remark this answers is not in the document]
hdamico: There is no need of adding that, status is a enum, so if it is nil you will get an error that nil is not included.

spec/factories/payloads/pull_request.rb
```diff
@@ -0,0 +1,19 @@
...
+    pull_request do
+      {
+        id: rand(1000..1100),
+        number: rand(1..100),
+        title: "Pull Request-#{rand(1..100)}",
+        node_id: 'MDExOlB1bGxSZXF1ZXN0Mjc5MTQ3NDM3',
+        state: 'open',
+        locked: 'false',
+        draft: 'false',
+        user: (attributes_for :user, id: rand(1..1000)).as_json
```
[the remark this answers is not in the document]
hdamico: I am still working on the factories, i will add Faker for this cases

spec/services/github_service.rb
hdamico: I will refactor the payload factories to have a sub factory with an after_create which will merge the requested payload. Here an example:
```
    factory :pull_request_payload_with_repository do
      after(:create) do |pull_request_payload|
        pull_request_payload.merge!((create :repository_payload))
      end
    end
```

hdamico: I will refactor the payload factories to have a sub factory with an after_create which will merge the requested payload. Here an example:
```
    factory :pull_request_payload_with_repository do
      after(:create) do |pull_request_payload|
        pull_request_payload.merge!((create :repository_payload))
      end
    end
```

app/models/events/pull_request.rb
hdamico: What do you mean? I didn't get the point

app/services/action_handler.rb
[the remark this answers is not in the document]
mrubi-rootstrap: Nop, no factory.

`PullRequest`, `ReviewComment` and `Review` classes all implement the methods

```
def resolve
  return unless handleable?

  handle_action
end

def handleable?
  ACTIONS.include?(@payload['action'])
end
```

All of these classes also inherit from `ActionHandler`.

I was suggesting to remove these methods from the subclasses and move them to `ActionHandler`:

```
class ActionHandler < BaseService
  # ...

  def resolve
    return unless handleable?
  
    handle_action
  end

  def handleable?
    ACTIONS.include?(@payload['action'])
  end
end
```
replacing the reference to the ACTIONS constant by a method named :handled_actions

```
class ActionHandler < BaseService
  # ...

  def resolve
    return unless handleable?
  
    handle_action
  end

  def handleable?
    handled_actions.include?(@payload['action'])
  end
end
```

```
class PullRequest < ActionHandler
  # ..

  def handled_actions
    ACTIONS
  end
end

# same for `ReviewComment` and `Review`
```

app/services/action_handlers/pull_request.rb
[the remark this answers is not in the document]
mrubi-rootstrap: Yes, the concern is related to the ownership of the `review_request` collection and in a more abstract sense with the question 'which object should modify the contents of a given collection (persisted or not)?'.

If the object PullRequest has the ownership of the `review_requests` collection it is an interesting idea that the only object modifying the collection be the PullRequest itself.
For that it would have a method call `create_review_request`.

If the object PullRequest does not have the ownership of that collection then the single point can be in the collection owner(s).

spec/services/github_service.rb
mrubi-rootstrap: Indeed, that is why I suggested using `build` instead. `:build` already has the meaning of building an object without persisting it into the database.

Somebody reading `build` in this method already knows that the `pull_request_payload` is not persisted without having to know the implementation details defined in a different place.

mrubi-rootstrap: How about having 2 different factories, one for a `pull_request_event_payload` and another one for a `pull_request_payload` (the PR object, not the PR event).

The same for `repository_event_payload` and `repository_payload`.

After all the definition of the Webhooks treats this concepts like composable objects.

This would allow to use composition of the events and objects like it does with other attributes:

```
# default pull request event that includes a default pull request and a default repository
build :pull_request_event_payload
```

```
# override the pull request event with a different repository
build : pull_request_event_payload, repository: (build :repository_payload)
```

```
# override of the pull request event with a different pull request and a different repository
build : pull_request_event_payload, pull_request: (build :pull_request_payload), repository: (build :repository_payload)
```

app/models/events/pull_request.rb
mrubi-rootstrap: I meant to declare not only the model relations but also the model attributes:

```
class Event < ApplicationRecord
  TYPES = %w[pull_request review review_comment].freeze

  attribute :data, default: {}
  attribute :handleable_type
  attribute :name
  attribute :type

  belongs_to :handleable, polymorphic: true, optional: true
  belongs_to :project
  validates :name, :data, presence: true
end
```

But as I said before probably it is related to a code style.

hdamico: This does not apply in Rails

hdamico:
Done! Ready for reviews!

app/services/github_service.rb
```diff
@@ -0,0 +1,48 @@
...
+      pj.description = repo['description']
+    end
+  end
+
+  def handle_event
+    event = Event.create!(project: @project, data: @payload, name: @event)
+    raise StandardError unless handleable_event?
+
+    @event_type = find_or_create_event_type
+    event.update!(handleable: find_or_create_event_type)
```
santiagovidal: Can't you use `@event_type` here instead of calling `find_or_create_event_type` again?

app/services/action_handlers/pull_request.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def closed
+      merged if @payload['pull_request']['merged'] == true
+      @event_type.closed!
+      @event_type.update!(closed_at: Time.current)
+    end
+
+    def review_request_removed
+      reviewer = find_user
```
santiagovidal: Why can't we use `find_or_create_user` here?

app/services/action_handler.rb
```diff
@@ -0,0 +1,23 @@
...
+  def call
+    return unless handleable?
+
+    handle_action
+  end
+
+  private
+
+  def find_user
```
santiagovidal: I would delete this method. You already have `find_or_create_user`.

Also, this method is only used by the `PullRequest` handler, so in any case, it should go there, right?

app/services/action_handler.rb
```diff
@@ -0,0 +1,23 @@
+class ActionHandler < BaseService
+  def initialize(payload:, event:, event_type:)
+    @payload = payload
+    @event = event
+    @event_type = event_type
```
santiagovidal: Does it make sense to rename this to something like `entity`? Because in fact it holds the actual entity we are going to update (PullRequest, Review, etc), right?

santiagovidal:
Shouldn't this go against `develop` instead of `master?

app/services/action_handler.rb
hdamico: Yes, you are right! Thanks!

app/services/action_handlers/pull_request.rb
hdamico: I thought of forcing to find a user, because in that part the user should already exist, but in the other hand, would be a problem if the user does not exist?

app/services/github_service.rb
hdamico: Yes, my bad!

app/services/action_handlers/pull_request.rb
santiagovidal: I guess if there's no user at this point, it could be a symptom of a bug in the code. Following that logic, it's fine to keep the method. I would move it to this class though, since it's only used here.

spec/controllers/webhook_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe WebhookController, type: :controller do
```
santiagovidal: We usually prefer request specs over controller specs. In this case we just have one endpoint so I wouldn't worry about it though.

app/services/action_handlers/pull_request.rb
hdamico: Ok!

PR #61, opened 2020-02-20
Implement the Rake task to process the ReviewTurnaround metrics

mrubi-rootstrap:
## What does this PR do?

Implements the Rake task

`rake metrics:process`

that creates a ProcessMetricsJob and evaluates it.

This PR does not implement the ProcessMetricsJob, it only implements the Rake task.

The implementation of the ProcessMetricsJob is to be done in a different PR.

### DIFF
--- a/app/jobs/process_metrics_job.rb
+++ b/app/jobs/process_metrics_job.rb
@@ -13,6 +13,7 @@ class ProcessMetricsJob < ApplicationJob
   #
   # It takes care of processing only the metrics that are pending to be
   # processed.
-  def perform()
+  def perform
+    # todo
   end
 end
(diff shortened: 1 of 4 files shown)


colembolos:
I have a question, why create a Job that does not have the implementation?. I think that we need first is to implement the service and then create the rake task.

colembolos:
One suggestion about the branch name, in the repo we have a guide, this particular branch sounds like `feature/create-process-metrics`, here the link: https://github.com/rootstrap/tech-guides/tree/master/git ;)

mrubi-rootstrap:
@colembolos 

To split the implementation into smaller tasks. This task only deals with the Rake task and uses only a job placeholder.

Regarding what to implement first, the rake task or the process, ideally it should be possible to do either. That is not always easy or possible but in this case I am trying a top-down approach.

mrubi-rootstrap:
@colembolos 

Oh, just I got the name! I thought that the `feature` part of the name was the placeholder for the ticket definition.

I will change the name. Thanks!

PR #67, opened 2020-02-21
Fix coverage

hdamico:
## What does this PR do?

Fixes coverage for code climate.


lib/simplecov_default_profile.rb
```diff
@@ -0,0 +1,11 @@
+SimpleCov.profiles.define 'default' do
```
santiagovidal: I think this should go in the `spec` folder, given that it's only usable by the test environment.

hdamico: Alright!

Issue #55, opened 2020-02-18
Process Review Turnaround Metrics in Metrics table


mrubi-rootstrap:
The metrics table keeps the value of
  - a metric
  - of an entity
  - at a given point in time or a given time interval

For example it keeps the collected values of the Review Turnaround metric value

To populate the metrics table a process runs collecting and calculating the values of interest for each given metric.

The process runs by demand as a standalone process or as a frequent cron job.

The process to collect metrics takes into account which are the previous values already collected to avoid collecting duplicated values (which might add corrupted values to the metric) and collects only the values that haven't been processed yet.

This task implements the metrics recollection process for a first concrete case, the Review Turnaround metric.

PR #77, opened 2020-03-03
Feature/collect review turnaround metrics 55

mrubi-rootstrap:
## What does this PR do?

Create the process that periodically collects the Review Turnaround metrics for projects and make the collected metrics available in the Metrics table.

Implements the story https://github.com/rootstrap/rs-code-review-metrics/issues/55

## Details

This PR does:

- define the rake task `rake metrics:process` that calls the job to generate the metrics
- define the ProcessMetricsJob that processes the events to generate the metrics
- define the service ReviewTurnaroundPerProjectProcessor that generates the ReviewTurnaround metrics for projects
- define the MetricsDefinition model that defines which metrics should be run when processing events

## Yet to be implemented

- It is missing the implementation ReviewTurnaroundPerUserProcessor and ReviewTurnaroundPerUserPerProjectProcessor. Let's create a new PR for those.


app/models/event.rb
```diff
@@ -23,5 +23,27 @@ class Event < ApplicationRecord
...
+  validates :handleable, presence: true, if: proc { |event| event.handled_type? }
+  validate :name_matches_handleable_type, if: proc { |event| handleable && event.handled_type? }
+
+  scope :received_after, ->(time) { time ? where('created_at > ?', time) : all }
+
+  ##
+  # Validate that the receiver name matches the class of the handleable Event.
+  # For example if handleable is a Events::PullRequest object the name is
+  # expected to be 'pull_request'
+  def name_matches_handleable_type
```
santib: I'd use a custom validator to prevent as much as possible having code in models https://guides.rubyonrails.org/active_record_validations.html#custom-validators

app/models/time_interval.rb
```diff
@@ -0,0 +1,20 @@
+##
+# An interval of time between two concrete points in time
+class TimeInterval
+  ##
+  # Creates a TimeInterval stargin at the given point in time with the given
```
santib: typo: starting

app/models/time_interval.rb
```diff
@@ -0,0 +1,20 @@
...
+  # duration expressed as a number of seconds
+  def initialize(starting_at:, duration:)
+    @starting_at = starting_at
+    @duration = duration
+    @closed_end_at = @starting_at + @duration - 0.01.seconds
```
santib: minor question: why not just `@end_at = @starting_at + @duration` and then in any place we could use `<` or `<=` ?

app/models/time_interval.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  attr_reader :starting_at, :duration
+
+  ##
+  # Returns true if the TimeInterval includes the given time.
+  def includes?(time)
+    time.between?(@starting_at, @closed_end_at)
```
santib: use attr readers (no `@`)

app/models/time_interval.rb
```diff
@@ -0,0 +1,20 @@
...
+  ##
+  # Creates a TimeInterval stargin at the given point in time with the given
+  # duration expressed as a number of seconds
+  def initialize(starting_at:, duration:)
+    @starting_at = starting_at
+    @duration = duration
+    @closed_end_at = @starting_at + @duration - 0.01.seconds
+  end
+
+  attr_reader :starting_at, :duration
```
santib: I'd put this above every method (including initialize) so it's easier to see

app/services/github_service.rb
```diff
@@ -41,6 +39,8 @@ def find_or_create_event_type
   end
 
   def handle_action
+    raise StandardError unless handleable_event?
```
santib: I thought that now we were "allowing" all the events even if some of them aren't handled now. What's the pointof raising an exception?

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
+module Metrics
+  ##
+  # Base class for classes that do the processing of a Metric from a
+  # collection of events.
+  class BaseMetricProcessor < BaseService
+    def initialize(events:, time_interval:)
```
santib: suggestion: use attr readers to avoid the `@` all over the place

app/services/github_service.rb
santib: for example this service is called by a job. If an exception is raised and the job is retried we'll be mistakenly creating multiple `Event`s

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
...
+      @events = events
+      @time_interval = time_interval
+    end
+
+    ##
+    # Processes the given events to generate the metrics.
+    # The given @events are expected to include all the events during the full
+    # time interval ( [@starting_at, @starting_at + @time_span) ).
+    def call
```
santib: this method can be confusing since we already have a `call` class method for the `BaseService`

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
...
+      process_events(events: @events, time_interval: @time_interval)
```
santib: you don't need to pass those params, they are sent in the initialization and can be used as instance variables

santib: merging this comment with the one above, I'd just remove the call method and make `process_events` public (without params)

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    private
+
+    ##
+    # Processes the given events to generate the review_turnaround metrics.
```
santib: why a "base metric processor" is talking about a particular metric such as "review_turnaround" ?

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      iterate_events(events: events, time_interval: time_interval, accumulators: accumulators)
+
+      update_metrics(time_interval: time_interval, accumulators: accumulators)
+    end
+
+    ##
+    # Processes all the given events that are included in the time interval
+    def iterate_events(events:, time_interval:, accumulators:)
+      events.each do |event|
```
santib: use `find_each` to prevent memory bloats https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

app/services/metrics/base_metric_processor.rb
santib: and as a general comment all over the file: use the attr_readers instead of passing the same params all over the place

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
...
+    #     {
+    #       accumulator_variable: value,
+    #       ...
+    #     }
+    # that is passed along to the :process_event method.
+    # The use of this accumulator allows to perform the calculation of many
+    # metrics in a single iteration of all the events of a given time interval.
+    # For example to calculate the review_turnaround for all the projects in a
+    # single iteration of the events instead of one interation for each project.
+    def create_accumulators
```
santib: why isn't this an instance variable?

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
...
+    ##
+    # MetricsProcessors are interested only in a some Events.
+    # If this method returns true the event is ignored.
+    #   event:, time_interval:
+    def skip_event?(*)
+      false
+    end
+
+    ##
+    # Sets the the value of the metric with the key (entity_key, metric_key).
```
santib: type: `the the`

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,84 @@
...
+    def update_metric(entity_key:, metric_key:, value:, value_timestamp:)
+      find_or_create_metric(entity_key: entity_key, metric_key: metric_key)
+        .update!(value: value, value_timestamp: value_timestamp)
+    end
+
+    ##
+    # Returns the Metric with the given (entity_key:, metric_key:) and evaluates
+    # the &block on it.
+    # If the Metric does not exist it creates it.
+    def find_or_create_metric(entity_key:, metric_key:, &block)
```
santib: where is this method called with a block?

app/services/metrics/review_turnaround_per_project_processor.rb
```diff
@@ -0,0 +1,120 @@
...
+  #     - project A, 2020-01-01 => 100
+  #     - project B, 2020-01-01 => 20
+  #     - project A, 2020-01-02 => 130
+  #     - project B, 2020-01-02 => 120
+  #
+  # For the same project and a monthly interval generates the following metrics
+  #
+  #     - project A, 2020-01-01 => 320
+  #     - project B, 2020-01-01 => 140
+  #     - project A, 2020-02-00 => 150
```
santib: here and below the date should be `2020-02-01` I guess

app/services/metrics/review_turnaround_per_project_processor.rb
```diff
@@ -0,0 +1,120 @@
...
+    #   to true if a previous Review event was the first Review to ignore all
+    #   the Review events after that one.
+    def create_accumulators
+      {
+        review_turnaround_per_project: Hash.new { |hash, key| hash[key] = [] },
+        pull_request_reviewed: {}
+      }
+    end
+
+    def process_event(event:, accumulators:)
```
santib: instead of passing over the accumulators I'd rather go with any of these options:
1. Having the accumulators be defined in the `initialize` method of each subclass and use it with an `attr_reader`
2. Define it as a method like
```
def accumulators
  @accumulators ||= {
     review_turnaround_per_project: Hash.new { |hash, key| hash[key] = [] },
     pull_request_reviewed: {}
  }
end

File: app/services/metrics/review_turnaround_per_project_processor.rb [hunk 17]
```diff
@@ -0,0 +1,120 @@
...
+      pull_request_id = event.data['pull_request']['id']
+
+      review_turnaround_value = review_turnaround_as_seconds(event: event)
+      accumulators[:review_turnaround_per_project][event.project.name] << review_turnaround_value
+      accumulators[:pull_request_reviewed][pull_request_id] = :reviewed
```
santib: processing 1 event record is enough to consider the PR as reviewed?

app/services/metrics/review_turnaround_per_project_processor.rb
```diff
@@ -0,0 +1,120 @@
...
+    ##
+    # Updates the metrics for all the project in the given review_turnaround_per_project
+    def update_metrics(time_interval:, accumulators:)
+      accumulators[:review_turnaround_per_project].each_pair do |project_id, values|
+        average_value = values.sum / values.size
+
+        update_metric(
+          entity_key: project_id,
+          metric_key: 'review_turnaround',
+          value: average_value,
```
santib: It'd be interesting to have the avg, min, max, median. Have you already discussed about this? otherwise we can start the conversation as something to do in the future

app/services/metrics/review_turnaround_per_project_processor.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    ##
+    # Returns the review turnaround value for the given event.
+    def review_turnaround_as_seconds(event:)
+      payload = event.data
+
+      # RFC should we move these times to the Event model?
+      reviewed_at = parse_time(payload['review']['submitted_at'])
+      review_requested_at = parse_time(payload['pull_request']['created_at'])
```
santib: more of a philosophical question, but what if I create the PR as a WIP or POC and I don't request for review until some time in the future? I guess there is nothing we can do without changing the dev process of all projects

app/services/metrics/review_turnaround_per_project_processor.rb
```diff
@@ -0,0 +1,120 @@
...
+    #       &&
+    #     - its PR has exactly one Review (the one created by this very same review_event)
+    #
+    # In all other cases the pull_request linked to this review_event was already
+    # reviewed and the review_event is ignored for the review_turnaround metric
+    def skip_event?(event:, time_interval:, accumulators:)
+      payload = event.data
+      event.name != 'review' ||
+        payload['action'] != 'submitted' ||
+        accumulators[:pull_request_reviewed].key?(payload['pull_request']['id']) ||
```
santib: so you are skipping the event if the `accumulators[:pull_request_reviewed]` already has the project key. But how did you know you needed to process the previous event and not this one? What guaranteed the order?

lib/tasks/metrics.rake
```diff
@@ -0,0 +1,12 @@
+##
+# The namespace used by the tasks that handles the Github event metrics.
+#
+# Call the task from the command line with
+#
+#       rake metrics:process
+namespace :metrics do
+  desc 'Processes the stored events and generates the all the metrics'
+  task process: :environment do
+    ProcessMetricsJob.new.perform
```
santib: ActiveJob is meant to use perform_now or perform_later, not instantiating the obeject or anything. `ProcessMetricsJob.perform_now`

spec/factories/events.rb
```diff
@@ -20,7 +20,56 @@
...
+    transient do
+      action {}
+      pull_request_payload {}
+    end
+
+    data { build :check_run_payload }
+
+    name { 'check_run' }
+
+    association :project, factory: :project
```
santib: is factory declaration needed?

spec/factories/events.rb
```diff
@@ -20,7 +20,56 @@
...
+
+    after(:build) do |event, factory|
+      event.data['action'] = factory.action if factory.action
+      event.data['pull_request'] = factory.pull_request_payload if factory.pull_request_payload
+    end
+
+    factory :event_pull_request do
+      transient do
+        created_at { Faker::Date.between(from: 1.month.ago, to: Date.yesterday) }
+        updated_at { Faker::Date.between(from: created_at, to: Date.yesterday) }
```
santib: `to: Time.current` maybe? to prevent the case `created_at == updated_at == Date.yesterday` ??

app/services/metrics/metrics_processor.rb
```diff
@@ -0,0 +1,73 @@
...
+    def events_to_process_for(metrics_definitions)
+      Event.received_after(minimum_time_among(metrics_definitions))
+           .order(:created_at)
+    end
+
+    ##
+    # Query all the metrics_definitions to get the minimum for the events to
+    # process
+    def minimum_time_among(metrics_definitions)
+      metrics_definitions.map(&:last_processed_event_time).compact.min
```
santib: can't you use `metrics_definitions.mininum(:last_processed_event_time)` or something similar?

app/services/metrics/metrics_processor.rb
```diff
@@ -0,0 +1,73 @@
...
+      metrics_definition.update!(last_processed_event_time: events.last.created_at)
+    end
+
+    ##
+    # Returns an array with the events to process.
+    # It polls the given metrics_definitions to know which events to query
+    # to avoid querying more events than needed.
+    def events_to_process_for(metrics_definitions)
+      Event.received_after(minimum_time_among(metrics_definitions))
+           .order(:created_at)
```
santib: is created_at enough? I mean with webhooks and jobs (with retries) I'm not sure if the order of creation in our DB will always match the order in which the events actually happened

app/services/metrics/metrics_processor.rb
```diff
@@ -0,0 +1,73 @@
...
+    def metrics_processor_for(metrics_definition)
+      metrics_definition.metrics_processor.constantize
+    end
+
+    ##
+    # Makes each metric defined to process all the events.
+    def process_all_metrics
+      metrics_definitions = all_metrics_definitions
+
+      events = events_to_process_for(metrics_definitions)
```
santib: shouldn't you grab the events to process in a metrics definition basis instead of doing that "min" processed at event logic?

app/services/metrics/metrics_processor.rb
```diff
@@ -0,0 +1,73 @@
+module Metrics
+  ##
+  # Does the processing of the Github events previously stored and leaves the
+  # generated metrics in the Metrics table.
+  #
+  # It takes care of processing only the metrics that are pending to be
+  # processed.
+  class MetricsProcessor < BaseService
```
santib: I find this class name confusing because there is also a  `BaseMetricProcessor` and the rest of the processor implementations

app/services/metrics/metrics_processor.rb
```diff
@@ -0,0 +1,73 @@
...
+        process(metrics_processor: metrics_processor,
+                events: events,
+                metrics_definition: metrics_definition)
+      end
+    end
+
+    ##
+    # Makes the given metric to process all the events.
+    def process(metrics_processor:, events:, metrics_definition:)
+      time_interval = metrics_definition.time_interval_starting_at(Time.zone.today)
```
santib: why starting at = today?

app/models/event.rb
mrubi-rootstrap: @santib  will do. Thanks!

app/jobs/process_metrics_job.rb
```diff
@@ -0,0 +1,12 @@
+##
+# This job processes the Github events previously stored, generates the
+# metrics and leaves them in a ready to use state.
+class ProcessMetricsJob < ApplicationJob
+  queue_as :default
```
santib: ActiveJob retries jobs by default. Should this job be retried in case of failure? 🤔 I think maybe not.. I'd use `discard_on StandadError` and log the error (or notify it to any exception service we might have)
https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-discard_on

app/validators/event_name_validator.rb
```diff
@@ -0,0 +1,11 @@
+##
+# Validates that the Event.name matches the class of the Event.handleable.
+# For example if Event.handleable is a Events::PullRequest object the Event.name
+# is expected to be 'pull_request'
+class EventNameValidator < ActiveModel::Validator
+  def validate(event)
+    return if event.name == event.handleable_type_name
+
+    event.errors[:event_name_validator] << 'Event.name must match Event.handleable type'
```
santib: Maybe the error should be attached to the `name` attribute of the event instead of `event_name_validator`

spec/support/null_processor.rb
```diff
@@ -0,0 +1,9 @@
+module Metrics
+  ##
+  # This Metric does nothing.
+  class NullProcessor < BaseMetricProcessor
```
santib: nice 👍

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,63 @@
+module Metrics
```
santib: I think we could add some tests for this file too, so we don't depend on tests for the specific implementations

app/validators/event_name_validator.rb
```diff
@@ -0,0 +1,11 @@
+##
+# Validates that the Event.name matches the class of the Event.handleable.
+# For example if Event.handleable is a Events::PullRequest object the Event.name
+# is expected to be 'pull_request'
+class EventNameValidator < ActiveModel::Validator
```
santib: It'd be good to add tests to this class too

app/services/metrics/base_metric_processor.rb
```diff
@@ -0,0 +1,63 @@
...
+      process_events
+    end
+
+    private
+
+    ##
+    # Processes the given events to generate the review_turnaround metrics.
+    def process_events
+      iterate_events
+      update_metrics
```
santib: it'd be good to define these 2 methods and raise an exception `Not Implemented` in the base metric processor

spec/support/metrics_specs_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+    GithubService.call(payload: payload, event: 'review')
+  end
+end
+
+##
+# Create a ReviewCommentEvent event as it would come from github for the testing
+# repository and a given PullRequest payload.
+# Return the created event payload.
+def create_review_comment_event(pull_request_event_payload:, repository_payload: nil)
+  repository_payload = test_repository_a_payload if repository_payload.nil?
```
santib: `repository_payload ||= test_repository_a_payload` ?

santib:
```suggestion
  repository_payload ||= test_repository_a_payload
```

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+
+    it 'it generates the metric for the second project' do
+      expect(second_metric).to have_attributes(
+        entity_key: 'Project B',
+        metric_key: 'review_turnaround',
+        value_timestamp: time_interval_to_process.starting_at
+      )
+
+      expect(second_metric_value_expressed_as_seconds).to be_within(1.second) .of(45.minutes)
```
santib: what do you think of freezing time instead of working with `be_within`?

Rails comes with helpers by default https://andycroll.com/ruby/replace-timecop-with-rails-time-helpers-in-rspec/

santib: or why is `be_within(1.second)` being used?

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+      expect(first_metric).to have_attributes(
+        entity_key: 'Project A',
+        metric_key: 'review_turnaround',
+        value_timestamp: time_interval_to_process.starting_at
+      )
+
+      expect(first_metric_value_expressed_as_seconds).to be_within(1.second) .of(20.minutes)
+    end
+
+    it 'it generates the metric for the second project' do
```
santib: I find these tests hard to follow. I need to scroll all over the place in order to understand what's going on 😕

app/services/metrics/base_metric_processor.rb
santib: nevermind

spec/support/metrics_specs_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+
+# Helpers methods for testing metrics
+
+def events_to_process
+  Event.all
+end
+
+# RFC where should these methods be defined? The make reference to :subject
+def process_all_events
+  subject.call(events: events_to_process,
```
santib: having the call defined here makes it a bit obscure IMHO

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+
+  context 'no review_turnaround events' do
+    describe 'when processing a collection containing no review_events' do
+      let(:create_test_events) do
+        pull_request_event_payload = create_pull_request_event(
+          action: 'opened',
+          created_at: Time.zone.parse('2020-01-01T15:10:00')
+        )
+
+        create_review_comment_event pull_request_event_payload: pull_request_event_payload
```
santib: use parenthesis in method calls, unless some specific cases but I think this is not the case. https://github.com/rootstrap/tech-guides/tree/master/ruby#method-invocation-parens

db/migrate/20200302120947_create_metrics_definitions.rb
```diff
@@ -0,0 +1,13 @@
+class CreateMetricsDefinitions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :metrics_definitions do |t|
+      t.string :metrics_name, null: false
+      t.string :time_interval, null: false # TODO: make it a Postgres enum
+      t.string :subject, null: false # TODO: make it a Postgres enum
+      t.string :metrics_processor, null: false
```
santib: should this be an enum too? Because it needs to match a class name, right?

spec/services/metrics/metrics_processor_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      it 'creates the concrete metrics processor to process the metrics' do
+        expect_any_instance_of(Metrics::NullProcessor).to receive(:call).once
+
+        subject.call
+      end
+
+      it 'updates the metric last_processed_event_time after processing the metric' do
+        subject.call
+
+        expect(metrics_definition.last_processed_event_time).to eql(processed_event_time)
```
santib: what about:

expect { subject.call }.to change { metrics_definition.last_processed_event_time }.from([something]).to(processed_event_time)

spec/services/metrics/metrics_processor_spec.rb
```diff
@@ -0,0 +1,57 @@
+require_relative '../../support/metrics_specs_helper'
+
+RSpec.describe Metrics::MetricsProcessor do
```
santib: This metrics processor works with multiple specific processors at the same time, maybe we can add tests for that case.

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+          entity_key: 'Project A',
+          metric_key: 'review_turnaround',
+          value_timestamp: time_interval_to_process.starting_at
+        )
+
+        expect(first_metric_value_expressed_as_seconds).to be_within(1.second) .of(15.minutes)
+      end
+
+      it 'does not create a new metric' do
+        expect(generated_metrics_count).to eq(1)
```
santib: `expect { Metric.count }.to not_change` ?

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+      end
+    end
+  end
+
+  context 'review turnaround definition' do
```
santib: ?? I don't understand what is this context about

app/services/metrics/review_turnaround_per_project_processor.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    ##
+    # Updates the metrics for all the project in the given review_turnaround_per_project
+    def update_metrics
+      review_turnaround_per_project.each_pair do |project_id, values|
+        average_value = values.sum / values.size
+
+        update_metric(
+          entity_key: project_id,
+          metric_key: 'review_turnaround',
```
santiagovidal: How do we know if this turnaround metric is for the week, month, day, etc?

app/services/metrics/metrics_definition_processor.rb
```diff
@@ -0,0 +1,70 @@
...
+        process(metrics_processor: metrics_processor,
+                events: events,
+                metrics_definition: metrics_definition)
+      end
+    end
+
+    ##
+    # Makes the given metric to process all the events.
+    def process(metrics_processor:, events:, metrics_definition:)
+      time_interval = metrics_definition.time_interval_starting_at(Time.zone.today)
```
santiagovidal: Isn't today the `ending_at` time?

spec/services/metrics/metrics_definition_processor_spec.rb
```diff
@@ -0,0 +1,57 @@
+require_relative '../../support/metrics_specs_helper'
+
+RSpec.describe Metrics::MetricsDefinitionProcessor do
+  subject { Metrics::MetricsDefinitionProcessor }
```
santiagovidal: `subject { described_class }`

spec/services/metrics/metrics_definition_processor_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      it 'updates the metric last_processed_event_time after processing the metric' do
+        subject.call
+
+        expect(metrics_definition.last_processed_event_time).to eql(processed_event_time)
+      end
+    end
+
+    context 'with events that had been processed already' do
```
santiagovidal: What's the difference between this context and the one above? You're just creating an `Event` in both

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+    (build :repository_payload, name: 'Project A')['repository']
+  end
+
+  let(:time_interval_to_process) do
+    TimeInterval.new(starting_at: Time.zone.parse('2020-01-01 00:00:00'),
+                     duration: 1.day)
+  end
+
+  context 'no review_turnaround events' do
+    describe 'when processing a collection containing no review_events' do
```
santiagovidal: I think this `describe` is redundant. The `context` already explains it.

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+  let(:test_repository_a_payload) do
+    (build :repository_payload, name: 'Project A')['repository']
+  end
+
+  let(:time_interval_to_process) do
+    TimeInterval.new(starting_at: Time.zone.parse('2020-01-01 00:00:00'),
+                     duration: 1.day)
+  end
+
+  context 'no review_turnaround events' do
```
santiagovidal: If possible use words like `when`, `with`, etc for contexts.

spec/services/metrics/project_review_turnaround_processor_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      end
+
+      it 'does not create a metric' do
+        expect(generated_metrics_count).to eq(0)
+      end
+    end
+  end
+
+  context 'metrics generation' do
+    describe 'if the metrics for the given time interval was not yet generated' do
```
santiagovidal: Same comment here.

app/models/metrics_definition.rb
```diff
@@ -15,14 +15,23 @@
...
+  # Returns the TimeInterval object, for example TimeIntervals::DailyInterval,
+  # for this MetricsDefinition
+  def time_period
+    @time_period ||= time_interval_to_time_period
+  end
+
+  ##
+  # Converts the receiver time_interval string to a TimeInterval object
+  def time_interval_to_time_period
+    TIME_INTERVALS.fetch(time_interval.to_sym)
```
horacio: Is it possible to have a sensible default here?

app/models/time_intervals/daily_interval.rb
```diff
@@ -0,0 +1,40 @@
...
+    ##
+    # Returns a TimeInterval during 1 day starting at the begining of the day of
+    # the given time
+    def containing(time)
+      TimeInterval.new(starting_at: to_begining_of_day(time), duration: 1.day).freeze
+    end
+
+    ##
+    # Returns the begining of the day of the given time
+    def to_begining_of_day(time)
```
santib:
```suggestion
    def to_beginning_of_day(time)
```

app/models/time_intervals/daily_interval.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    ##
+    # Returns a TimeInterval during 1 day starting at the begining of the day of
+    # the given time
+    def containing(time)
+      TimeInterval.new(starting_at: to_begining_of_day(time), duration: 1.day).freeze
+    end
+
+    ##
+    # Returns the begining of the day of the given time
```
santib:
```suggestion
    # Returns the beginning of the day of the given time
```

santib: and this same typo in other places

app/services/metrics/metrics_definition_processor.rb
```diff
@@ -44,7 +53,7 @@ def process_each(metrics_definition:, events:)
     # to avoid querying more events than needed.
     def events_to_process_for_all(metrics_definitions)
       Event.received_after(oldest_event_time_to_process_all(metrics_definitions))
-           .order(:created_at)
+           .order(:occured_at)
```
horacio:
```suggestion
           .order(:occurred_at)
```

app/models/metrics_definition.rb
```diff
@@ -29,9 +29,11 @@ def time_period
     @time_period ||= time_interval_to_time_period
   end
 
+  private
+
   ##
   # Converts the receiver time_interval string to a TimeInterval object
   def time_interval_to_time_period
-    TIME_INTERVALS.fetch(time_interval.to_sym)
+    time_interval ? TIME_INTERVALS.fetch(time_interval.to_sym) : nil
```
horacio:
```suggestion
    TIME_INTERVALS[time_interval.to_sym]
```

app/models/events/pull_request.rb
```diff
@@ -43,5 +45,15 @@ class PullRequest < ApplicationRecord
               :locked,
               inclusion: { in: [true, false] }
     validates :github_id, uniqueness: true
+
+    ##
+    # Returns the number of reviews this pull_request has
+    delegate :count, to: :reviews, prefix: true
```
hdamico: To get the number of `reviews` of a `pull_request` we need an instance of a `pull_request` and then `instance_of_pr.reviews.count`, in this case we allow doing something like this `Events::PullRequest.count_reviews` which will give us the amount of reviews that exists in the database, that will be the same as `Events::Reviews.count`.

app/models/event.rb
```diff
@@ -23,5 +25,24 @@ class Event < ApplicationRecord
 
   belongs_to :handleable, polymorphic: true, optional: true
   belongs_to :project
-  validates :name, :data, presence: true
+
+  validates :name, presence: true
+  validates :data, presence: true
+  validates :handleable, presence: true, if: proc { |event| event.handled_type? }
+  validates_with EventNameValidator, if: proc { |event| handleable && event.handled_type? }
```
hdamico: Could you explain to me the purpose of this validation?

app/models/events/event_record.rb
```diff
@@ -0,0 +1,9 @@
+module Events
+  ##
+  # This Module adds common method to all ApplicationRecord modeling Events
+  module EventRecord
+    def event_name
+      self.class.name.demodulize.underscore
+    end
+  end
+end
```
hdamico: Wouldn't be better to create a helper to do this?

app/models/events/review.rb
```diff
@@ -35,5 +37,29 @@ class Review < ApplicationRecord
...
+      review_payload = payload.fetch('review')
+      case event_action
+      when 'submitted'
+        parse_time(review_payload.fetch('submitted_at'))
+      when 'edited'
+        parse_time(review_payload.fetch('edited_at'))
+      when 'dismissed'
+        parse_time(review_payload.fetch('dismissed_at'))
+      else
+        raise "Uknown action '#{event_action}'"
```
hdamico: We should not raise errors. As Santi told, this is a job, whenever it raises an error it will retry, which will come to an infinity loop.

hdamico: I believe that this a GitHub logic, I would take it off from the model and put it in a service

app/models/events/review.rb
```diff
@@ -35,5 +37,29 @@ class Review < ApplicationRecord
...
+      end
+    end
+
+    ##
+    # Parses the given time_string and returns a TimeWithZone object
+    def parse_time(time_string)
+      Time.zone.parse(time_string)
+    end
```
hdamico: I would put this on a helper

app/models/events/pull_request.rb
```diff
@@ -43,5 +45,15 @@ class PullRequest < ApplicationRecord
...
+
+    ##
+    # Returns the occurrence time of the 'pull_request' event.
+    def occurence_time(payload:)
```
santiagovidal: I don't think the model should have this code.

app/models/time_intervals/daily_interval.rb
```diff
@@ -0,0 +1,40 @@
...
+    ##
+    # Returns a TimeInterval during 1 day starting at the beginning of the day of
+    # the given time
+    def containing(time)
+      TimeInterval.new(starting_at: to_beginning_of_day(time), duration: 1.day).freeze
+    end
+
+    ##
+    # Returns the beginning of the day of the given time
+    def to_beginning_of_day(time)
```
santiagovidal: You can use the method `beginning_of_day` already provided by Rails.

app/services/metrics/base_metric_processor.rb
mrubi-rootstrap: The meaning of NotImplemented in Ruby is different from Python and other languages: [https://ruby-doc.org/core-2.5.0/NotImplementedError.html](https://ruby-doc.org/core-2.5.0/NotImplementedError.html).

Also [this](https://ruby-doc.org/core-2.5.0/NotImplementedError.html) style guide suggest to prefer duck tape.

spec/support/metrics_specs_helper.rb
mrubi-rootstrap: Refactored. Thanks!

spec/services/metrics/project_review_turnaround_processor_spec.rb
mrubi-rootstrap: Changed it. Thanks!

spec/services/metrics/project_review_turnaround_processor_spec.rb
mrubi-rootstrap: Fixed it. Thanks!

spec/services/metrics/metrics_processor_spec.rb
mrubi-rootstrap: Done. Thanks!

app/models/events/event_record.rb
mrubi-rootstrap: Done it. Thanks!

app/models/events/review.rb
mrubi-rootstrap: I think we should configure the maximum number of retries to avoid infinite loops.
I don't think it is possible to rely on a Job not raising errors during its execution. Runtime errors can be raised for reasons not in control of the Job (find! or create! failures, lost connections to databases or queues, unexpected or inconsistent data in a records).

mrubi-rootstrap: Done it. Thanks!

app/models/events/review.rb
mrubi-rootstrap: Done it. Thanks!

spec/support/metrics_specs_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+  first_metric.value.seconds
+end
+
+def second_metric
+  Metric.second
+end
+
+def second_metric_value_expressed_as_seconds
+  second_metric.value.seconds
+end
```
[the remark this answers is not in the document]
mrubi-rootstrap: Refactored to let statements. Thanks!

app/services/queries/reviews_turnaround_per_project_query.rb
```diff
@@ -0,0 +1,60 @@
+module Queries
+  class ReviewsTurnaroundPerProjectQuery
+    attr_reader :time_interval
+
+    def initialize(time_interval:)
+      @time_interval = time_interval
+    end
+
+    def last_review_created_at
+      Events::Review.order(:created_at).pluck(:created_at).last
```
santiagovidal: I think it's better to do `.first.created_at` instead of `pluck` because at some point there will be many reviews, and will take more memory space than just one object.

mrubi-rootstrap: Fixed it. Thanks!

spec/models/event_spec.rb
```diff
@@ -21,17 +21,19 @@
...
+    context 'of an event handled type' do
+      it { should validate_presence_of(:handleable) }
 
-    it 'is not valid without name' do
-      subject.name = nil
-      expect(subject).to_not be_valid
+      it 'should validate the name to match the event type' do
+        subject = build(:event_pull_request, name: 'review')
+        expect(subject).not_to be_valid
+      end
```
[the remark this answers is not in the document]
mrubi-rootstrap: Fixed it. Thanks!

lib/tasks/metrics.rake
```diff
@@ -0,0 +1,12 @@
+##
+# The namespace used by the tasks that handles the Github event metrics.
+#
+# Call the task from the command line with
+#
+#       rake metrics:process
+namespace :metrics do
+  desc 'Processes the stored events and generates the all the metrics'
+  task process: :environment do
+    ProcessMetricsJob.new.perform_now
```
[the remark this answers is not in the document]
mrubi-rootstrap: Fixed it. Thanks!

app/models/event.rb
```diff
@@ -23,5 +23,23 @@ class Event < ApplicationRecord
 
   belongs_to :handleable, polymorphic: true, optional: true
   belongs_to :project
-  validates :name, :data, presence: true
+
+  validates :name, presence: true
+  validates :data, presence: true
```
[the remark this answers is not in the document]
mrubi-rootstrap: Fixed it. Thanks!

app/services/github_service.rb
```diff
@@ -25,11 +25,9 @@ def build_project
   end
 
   def handle_event
-    event = Event.create!(project: @project, data: @payload, name: @event)
-    raise StandardError unless handleable_event?
+    @entity = find_or_create_event_type if handleable_event?
```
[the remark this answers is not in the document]
mrubi-rootstrap: Done. Thanks!

PR #80, opened 2020-03-03
Generate basic views

### LINKED ISSUE #79: Generate basic views
## Description

Generate views with sprockets instead of ActiveAdmin to add code flexibility.

hdamico:
## What does this PR do?

Generates basics views for metrics

Screen Shot 2020-03-10 at 11 20 26

Resolves #79


matiasmansilla1989:
Adding @bocha, he can provide good feedback about frontend stuff.

app/views/shared/_header_menu.html.erb
```diff
@@ -0,0 +1,6 @@
+<ul class="nav nav-background nav-txt-color p-2">
+  <a class="navbar-brand nav-txt-color" style="color: white" href="/">Github Analyzer</a>
```
bocha: Let's avoid inline styles in the `HTML` markup, unless they are critical ones loaded in the layout to improve performance over unused ones above the fold.

What do you think of creating a `CSS` file called `_nav_bar.scss` (or similar) inside `stylesheets/components/`, where you can declare all styles related to this Nav Bar?

app/views/shared/_header_menu.html.erb
```diff
@@ -0,0 +1,6 @@
+<ul class="nav nav-background nav-txt-color p-2">
```
bocha: Given that we have only one list item inside here... what do you think of using a `nav` tag instead of `ul`? 🤔 
That will improve tremendously the semantic relevance of the markup!
We can even keep this `ul > li` estructure, but housed inside a `nav` 😉

bocha:
@hdamico Can we have a screenshot or animated gif of the page with these changes?
In that way, I can also review and provide feedback on a design perspective 🙂

app/views/metrics/review_turnaround.erb
```diff
@@ -0,0 +1,3 @@
+<div class='p-3'>
+  <%= line_chart Event.group(:created_at).count %>
```
santiagovidal: This is just as an example right? I'd add a comment here to clarify that this should be removed.

app/assets/stylesheets/shared/header_menu.scss
```diff
@@ -0,0 +1,12 @@
+$nav-background: #ab0404;
+$nav-border: 1px solid rgb(75, 75, 75);
```
bocha: Let's try to be more descriptive with these variable names since they could be anything, not necessary only colors. What do you think about: `$nav-bg-color` and `$nav-border-color`?

bocha: Also, is OK to leave them here for now since they are only a couple, but in the future, bear in mind to create an exclusive stylesheet for them, like `stylesheets/abstracts/_variables.scss` 😬

hdamico: Alright! Will keep it in mind

PR #86, opened 2020-03-10
Delete team from CODEOWNERS

hdamico:
## What does this PR do?

Deletes the team from CODEOWNERS.
After reading the documentation I realize that the only way to make it work with team you have to add this repository to the team and give read access to it, otherwise this won't work, so I kept only the user, in this way there are no permission problems.


mrubi-rootstrap:
Should we create a code_review_metrics team and grant permissions to that team?

hdamico:
> Should we create a code_review_metrics team and grant permissions to that team?

There is no need of that. Using team namespace, you can only refer to a hole team and not to a specific user. Once the user leaves the organization he is not longer a code owner.

PR #87, opened 2020-03-10
Fix flaky tests

### LINKED ISSUE #85: Fix Flaky tests
## Description

Currently, we are having lots of flaky tests because of bad Faker gem implementation. We should only use it on non-important fields.

## Things to do

* Replace `id` in every factory to a sequence.
* Same with body and important fields.

hdamico:
## What does this PR do?

Fixes flaky tests. Currently, we are generating random data on key attributes, for example, `id` and `body`, which are causing random failures, so by adding `sequence` instead of `Faker` we avoid this.

Resolves #85

### DIFF
--- a/spec/factories/events.rb
+++ b/spec/factories/events.rb
@@ -19,6 +19,8 @@
 #
 
 FactoryBot.define do
+  sequence(:body) { |n| "Please change line #{n}" }
+
   factory :event do
     name { %w[pull_request review review_comment].sample }
     data { Faker::Json.shallow_json(width: 3) }
--- a/spec/factories/events/review_comments.rb
+++ b/spec/factories/events/review_comments.rb
@@ -23,9 +23,11 @@
 #
 
 FactoryBot.define do
+  sequence(:review_comment_id, 100)
+
   factory :review_comment, class: Events::ReviewComment do
-    github_id { Faker::Number.unique.number(digits: 4) }
-    body { Faker::Quote.singular_siegler }
+    github_id { generate(:review_comment_id) }
+    body { generate(:body) }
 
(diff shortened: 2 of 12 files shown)


spec/factories/events/review_comments.rb
```diff
@@ -23,9 +23,11 @@
 #
 
 FactoryBot.define do
+  sequence(:review_comment_id, 100)
+
   factory :review_comment, class: Events::ReviewComment do
-    github_id { Faker::Number.unique.number(digits: 4) }
-    body { Faker::Quote.singular_siegler }
+    github_id { generate(:review_comment_id) }
+    body { generate(:body) }
```
horacio: Here you can use the sequence as an implicit attribute, like so

```suggestion
    body
```

Same for other occurrences in this PR.

spec/factories/events.rb
```diff
@@ -19,6 +19,8 @@
 #
 
 FactoryBot.define do
+  sequence(:body) { |n| "Please change line #{n}" }
```
horacio: What do you think about renaming this sequence to something like `review_body`?

hdamico: This sequence is being used by different factories (review_payload, review_comment_payload and review_comment), so I put it in Event factory, which comes to be a common place in sense of hierarchy.
Should I create a sequence for each one?

horacio: Ahhh, I see. You're right. Sorry, now I remember we talked about this yesterday.

Maybe for this attribute we do can use Faker, right?

horacio: Solved in an offline conv.

PR #88, opened 2020-03-10
Fix code duplication

hdamico:
## What does this PR do?

Fixes code duplication which has been reported by CodeClimate.


app/services/action_handler.rb
```diff
@@ -1,4 +1,6 @@
 class ActionHandler < BaseService
+  private_constant :ACTIONS
```
horacio: This will break the tests. :(

This change doesn't have the intended effect. We need to keep those `private_constant`s in place.

PR #97, opened 2020-03-18
Hotfix jobs retries

hdamico:
## What does this PR do?

Fixes loop retries on Sidekiq. We were raising error if the event was not handleable, which came to a retry.


db/migrate/20200312144232_rename_value_on_pr_state.rb
```diff
@@ -1,13 +0,0 @@
-class RenameValueOnPrState < ActiveRecord::Migration[6.0]
```
santib: Removing a migration? Why?

santiagovidal: I would avoid modifying/deleting migrations that were already run in an environment. In this case maybe it's better to create a new migration to rename the column again.

app/models/events/pull_request.rb
```diff
@@ -24,7 +24,7 @@
 #
 module Events
   class PullRequest < ApplicationRecord
-    enum state: { opened: 'opened', closed: 'closed' }
```
horacio: Why are we rolling back on this?

db/migrate/20200312144232_rename_value_on_pr_state.rb
hdamico: Is going to be reverted. The structure is changed so when it gets on Heroku it will execute the migrations again and it will revert that one.

app/models/events/pull_request.rb
hdamico: Because we receive `open` instead of `opened` from GitHub, and it detects it as an invalid state.

db/migrate/20200312144232_rename_value_on_pr_state.rb
hdamico: Ok! Thought of not repeating code

PR #103, opened 2020-03-26
Little refactor

hdamico:
## What does this PR do?

Organizes the code and gets rid of some validations


app/helpers/project_builder.rb
```diff
@@ -0,0 +1,8 @@
+module ProjectBuilder
```
santib, line 1: My personal experience is that these techniques using modules to share behavior is hard to maintain/extend. They can quickly grow out of control, start having dependencies between modules and bidirectional access with the classes that includes this modules. Also it's hard to follow the code since you can call methods without being explicit on where that method is defined. My recommendation is just using other POROs if you want to encapsulate behavior

app/models/event.rb
```diff
@@ -27,19 +27,4 @@ class Event < ApplicationRecord
   belongs_to :project
 
   validates :name, :data, presence: true
-  validates :handleable, presence: true, if: proc { |event| event.handled_type? }
-  validates_with EventNameValidator, if: proc { |event| handleable && event.handled_type? }
-  validates :occurred_at, presence: true, if: proc { |event| event.handled_type? }
```
santib, line 32: why are we removing validations?

app/helpers/project_builder.rb
santib, line 1: Moreover the `helpers` folder in Rails has a meaning and is to use methods directly in the `views`. Which BTW is pattern I think should be use with caution too.

app/models/event.rb
hdamico: The first validation is in the github service already, we are duplicating code and making the validation twice.
The second one I believe is not necessary at all (correct me if I'm wrong)
The third one, the same as the first one.

santib: @mrubi-rootstrap might know better. Regarding the first and third validation, I get your point but remember that if in the future there is another piece of code creating/updating events then we'll be missing these validations

hdamico: So would it be better to validate it only in the model? I would like to keep those only in one place.

santib: ideally we would do business validations outside models and integrity validations in the models. What do you think these 2 are business or integrity validations?

app/services/github_service.rb
```diff
@@ -13,41 +13,37 @@ def call
...
-    Event.create!(project: @project,
-                  data: @payload,
-                  name: @event,
-                  handleable: @entity,
-                  occurred_at: occurence_time(event_name: @event, payload: @payload))
+    project = Projects::Builder.fetch_or_create(@payload['repository'])
+    event = Event.new(project: project, data: @payload, name: @event)
+    if handleable_event?
+      @entity = find_or_create_event_type
+      event.tap do |evnt|
```
santib, line 25: I think you don't need to use `tap` here

app/services/github_service.rb
```diff
@@ -13,41 +13,37 @@ def call
...
+        evnt.assign_attributes(handleable: @entity, occurred_at: occurence_time)
+        evnt.save!
+      end
+    else
+      event.tap(&:save!)
```
santib, line 30: same here

app/models/event.rb
hdamico: I believe is most likely to be a business validation rather than an integrity validation.

app/services/github_service.rb
hdamico: I need to return the `event` here, and `.save` does not return the object, instead it returns `true` or `false`.

santib: you still have event, this would be the same
```
  event.assign_attributes(handleable: @entity, occurred_at: occurence_time)
  event.save!
  event
```

hdamico: Alright

config/rails_best_practices.yml
```diff
@@ -28,7 +28,8 @@ RemoveUnusedMethodsInHelpersCheck: { except_methods: [
   'GithubEventPayloadHelper#occurence_time',
```
santiagovidal, line 28: Can you remove this and any other method exceptions that are no longer valid?

hdamico: Yep!

PR #108, opened 2020-04-03
Associaton between user and project

hvilloria:
## What does this PR do?

* Adds a references from project or user in metric model
* Removes metric_key from metric model since the metric_definition has which metric is being processed.
* changes some tests to handle these new attributes in metric model.
* changes the metric factory with its association


hdamico:
Please relate this PR with an issue.

PR #110, opened 2020-04-06
Review turnaround process fix

### LINKED ISSUE #104: Fix Review Turnaround query
## Description

Currently, the query that we have to generate the Review Turnaround metric is incorrect.

### Things to do

- [ ] - Re-build the query.
- [ ] - Fix tests

hdamico:
## What does this PR do?

1. Fixes review turnaround workflow and query.
2. Creates a relation between `User` and `Project` (`UserProject`).
3. Creates a relation between `PullRequest` and `Owner` (`User`).
4. Rewrites Review Turnaround query and fix tests.
5. Creates a relation between `ReviewRequest` and `Review`.
6. Organizes code.

Resolves #104


spec/services/metrics/review_turnaround_per_user_processor_spec.rb
```diff
@@ -32,65 +31,61 @@
...
-          created_at: Time.zone.parse('2020-01-01T15:10:00')
+          action: 'opened'
         )
 
+        create_review_request_event(pull_request_event_payload: pull_request_event_payload)
+
         create_review_event pull_request_event_payload: pull_request_event_payload,
                             action: 'submitted',
-                            submitted_at: Time.zone.parse('2020-01-01T15:30:00')
+                            submitted_at: (Time.zone.now - 30.minutes).change(usec: 0)
```
hvilloria, line 41: 👏

app/models/events/review.rb
```diff
@@ -37,6 +39,7 @@ class Review < ApplicationRecord
                        inverse_of: :owned_reviews
     belongs_to :pull_request, class_name: 'Events::PullRequest',
                               inverse_of: :reviews
+    belongs_to :review_request
```
santiagovidal, line 42: This is optional right? What if I review a PR without being requested?

app/models/user.rb
```diff
@@ -34,6 +34,12 @@ class User < ApplicationRecord
            foreign_key: :reviewer_id,
            dependent: :destroy,
            inverse_of: :reviewer
+  has_many :users_projects, dependent: :destroy
+  has_many :projects, through: :users_projects
+  has_many :pull_requests,
```
santiagovidal, line 39: Can we rename this to `created_pull_requests`? It would better reflect the nature of the relationship

app/services/queries/reviews_turnaround_per_user_query.rb
```diff
@@ -0,0 +1,28 @@
...
+    private
+
+    def query
+      Events::Review.joins(:review_request, pull_request: :project).includes(:review_request)
+                    .where(opened_at: @time_interval.starting_at..@time_interval.ending_at)
+    end
+
+    def run
+      query.collect do |review|
+        { user_id: review.owner.id,
```
santiagovidal, line 22: We want to track the review turnaround on each specific project, right? So this should be the id of the `UserProject` entity

app/models/events/review.rb
hdamico: AFAIK once you add a review, it adds you in the requested reviews list, so it triggers two events. The only problem that I see would be if the review creation goes before the review request creation. Maybe we should prioritize the jobs?

app/models/user.rb
hdamico: Alright!

app/services/queries/reviews_turnaround_per_user_query.rb
hdamico: Yes! My bad, already fixed.

app/services/event_builder.rb
```diff
@@ -13,4 +13,19 @@ def call
...
+    if review_requests.empty?
+      ReviewRequest.create!(owner: pull_request.owner,
+                            reviewer_id: reviewer_id,
+                            pull_request: pull_request)
+    else
+      review_requests.order(:created_at).first
+    end
+  end
+
+  def relate_user_project(user, project)
```
santiagovidal, line 28: Are you using this method?

app/services/queries/reviews_turnaround_per_user_project_query.rb
```diff
@@ -0,0 +1,32 @@
...
+    private
+
+    def query
+      Events::Review.joins(:review_request, pull_request: :project).includes(:review_request)
+                    .where(opened_at: @time_interval.starting_at..@time_interval.ending_at)
+    end
+
+    def run
+      query.collect do |review|
+        { entity: find_user_project(review.owner, review.pull_request.project),
```
santiagovidal, line 22: We have an N+1 here, but I'm not sure how to fix it 🤔

app/models/events/review.rb
santiagovidal: I don't think prioritizing the jobs is guaranteed to fix the problem. Would there be a problem if we leave the association as optional? Do we use it for something?

app/services/event_builder.rb
hdamico: No really, thanks

app/services/queries/reviews_turnaround_per_user_project_query.rb
hdamico: I set a breakpoint before the `run` method, and put `ActiveRecord::Base.logger = Logger.new(STDOUT)` to get the output of the query and I got just one query.

app/models/events/pull_request.rb
```diff
@@ -16,23 +16,28 @@
...
 #
+#  fk_rails_...  (owner_id => users.id)
 #  fk_rails_...  (project_id => projects.id)
 #
 module Events
   class PullRequest < ApplicationRecord
     enum state: { open: 'open', closed: 'closed' }
 
     belongs_to :project, inverse_of: :pull_requests
+    belongs_to :owner, class_name: 'User', foreign_key: :owner_id,
```
santib, line 39: question, what happens with existing records that don't have a `owner_id`? Are they going to be considered "invalid" for the rest of their lives?

hdamico: At some point we will need to re process all the events again because of the amount of changes after the deployment. But yes, you have a point here, the migration will fail because of existing records without the owner_id, at first we should have this field with a default on the db side, and delete this constraint when we have re processed all the events.

app/services/event_builders/review.rb
```diff
@@ -5,13 +5,22 @@ class Review < EventBuilder
...
     end
+
+    def assign_attrs(review, review_data)
+      review.owner = find_or_create_user(review_data['user'])
+      review.pull_request = find_pull_request
+
+      owner = review.owner
+      pull_request = review.pull_request
+      review.review_request = find_or_create_review_request(pull_request, owner.id)
+      owner.projects << pull_request.project
```
santib, line 23: question, the code above is "assigning" but not persisting. But this line is actually persisting the change, can it have any negative impact like leaving system in an invalid state if something fails?

app/services/metrics/review_turnaround_per_user_project_processor.rb
```diff
@@ -0,0 +1,17 @@
+module Metrics
+  class ReviewTurnaroundPerUserProjectProcessor < MetricProcessor
+    private
+
+    def process
+      review_turnaround_average.each do |review_turnaround|
```
santib, line 6: how many records will we have here? should we use `find_each` instead? https://apidock.com/rails/ActiveRecord/Batches/find_each

app/services/queries/reviews_turnaround_per_user_project_query.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    private
+
+    def query
+      Events::Review.joins(:review_request, pull_request: :project).includes(:review_request)
+                    .where(opened_at: @time_interval.starting_at..@time_interval.ending_at)
+    end
+
+    def run
+      query.collect do |review|
```
santib, line 21: `collect` will be fetching all the records in memory at once, never tried it but maybe the `lazy` here https://collectiveidea.com/blog/archives/2015/03/05/optimizing-rails-for-memory-usage-part-3-pluck-and-database-laziness is what you need?

santib: https://ruby-doc.org/core-2.5.0/Enumerator/Lazy.html

app/services/queries/reviews_turnaround_per_user_project_query.rb
santib: but if you are fetching the owner for every record returned, how can it not be triggering N queries?

app/services/queries/reviews_turnaround_per_user_project_query.rb
```diff
@@ -0,0 +1,32 @@
...
+        { entity: find_user_project(review.owner, review.pull_request.project),
+          turnaround: [review.review_request.created_at.to_i,
+                       review.opened_at.to_i].reduce(:-) }
```
santib, line 24: alternative
```suggestion
        {
          entity: find_user_project(review.owner, review.pull_request.project),
          turnaround: review.review_request.created_at.to_i - review.opened_at.to_i
        }
```

hdamico: I did it in that way, but got a line length error from Rubocop...

app/services/queries/reviews_turnaround_per_user_project_query.rb
hdamico: Hey! Thank you for making me realize of that! I am going to try that approach.

app/services/event_builders/review.rb
hdamico: I got an error from reek that I was doing `review.owner` and `review.pull_request` twice, so assigning solve that. I will add an exception on reek i guess

app/services/queries/reviews_turnaround_per_user_project_query.rb
hdamico: By doing the join I am getting all the owners. What I am missing is the owners in the join.

app/models/events/review.rb
hdamico: I found it necessary when it comes to find when the review was asked. For example: `review.created_at - review.review_requested.created_at` to get how much did the user last to make the review.

app/services/queries/reviews_turnaround_per_user_project_query.rb
hdamico: I got an approach: `review.review_request.created_at.to_i.-(review.opened_at.to_i)` then with the parenthesis I can put it in two lines.

app/models/events/review.rb
santiagovidal: If there's no review request, then there's no review turnaround to measure, so we can ignore those reviews while calculating the metric.

PR #111, opened 2020-04-07
Proposal for metrics services

hvilloria:
## What does this PR do?
  
This is a proposal for metrics services, the main idea is doing a little refactor by steps. In this case processors and metrics are separated, in this way we can have inside a metric service all the metrics that we will graph, besides having a more semantic distribution for the metrics and all the processors it needs.

Pr for issue: #109


spec/services/metrics/time_intervals/daily_metrics_spec.rb
```diff
@@ -1,6 +1,6 @@
 require_relative '../../../support/helpers/metrics_specs_helper'
```
hdamico, line 1: Could you change this like the others?

hvilloria: sure!

PR #117, opened 2020-04-09
Memory performance

hdamico:
## What does this PR do?

Improves Review Turnaround process. We were iterating the query result twice, with this improve, we iterate once, in batches and lazy, which comes in a huge memory improve in large scales.


app/services/queries/reviews_turnaround_per_user_project_query.rb
```diff
@@ -7,26 +7,15 @@ def initialize(time_interval:)
...
+      execute
     end
 
     private
 
-    def query
+    def execute
       Events::Review.joins(:review_request, owner: :users_projects, pull_request: :project)
                     .includes(:review_request)
                     .where(opened_at: @time_interval.starting_at..@time_interval.ending_at)
```
santib, line 18: the weird part is that this query object know how is it going to be used, I mean because of the `includes` and `joins`

hdamico: Yep, you are right. I will make that change on other PR

PR #119, opened 2020-04-13
Move GithubEventPayloadHelper

### LINKED ISSUE #112: Move GithubEventPayloadHelper
## Description

Move `GithubEventPayloadHelper` out of Helpers folder.
Helpers folders should be for views helpers, so let's use Plain Old Ruby Objects to encapsulate this methods.

hdamico:
## What does this PR do?

Moves `GithubEventPayloadHelper` from helpers folder to a PORO.

Resolves #112


app/services/payload/parser.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      def review_ocurrence_time(payload)
+        event_action = payload['action']
+        review_payload = payload['review']
+        if event_action == 'submitted'
+          Time.zone.parse(review_payload['submitted_at'])
+        elsif event_action == 'edited'
+          Time.zone.parse(review_payload['edited_at'])
+        elsif event_action == 'dismissed'
+          Time.zone.parse(review_payload['dismissed_at'])
```
horacio: WDYT about

```ruby
def review_ocurrence_time(payload)
  occurrence_time = payload['review']["#{payload['action']}_at"]

  Time.zone.parse(occurrence_time)
end
```

hdamico: Yep, great approach, thanks!

app/services/payload/parser.rb
```diff
@@ -0,0 +1,18 @@
+module Payload
+  class Parser
+    class << self
+      def pull_request_ocurrence_time(payload)
```
santib, line 4: the 3 methods receive `payload` as a param. That looks to me like `payload` could be part of the initialization of the service

PR #124, opened 2020-04-15
Name and interval for Metric model

### LINKED ISSUE #125: Add interval and name in metric model
## Description

Add interval and name in metric model.
Interval should be an enum with values:  'daily', 'weekly', 'monthly' and 'all_times'.
Name should be enum aswell with values: 'review_turnaround'

hdamico:
## What does this PR do?

Adds interval and name for Metric model.
Interval is an enum as well as name, the possible values for intervals are 'daily', 'weekly', 'monthly' and 'all_times', and for name is 'review_turnaround'.

Resolves #125


app/models/metric.rb
```diff
@@ -22,6 +24,12 @@
...
 
 class Metric < ApplicationRecord
+  enum interval: { daily: 'daily', weekly: 'weekly', monthly: 'monthly', all_times: 'all_times' }
+  enum name: { review_turnaround: 'review_turnaround' }
+
   belongs_to :metrics_definition, inverse_of: :metrics
   belongs_to :ownable, polymorphic: true
+
+  validates :interval, inclusion: { in: intervals.keys }
+  validates :name, inclusion: { in: names.keys }
```
horacio, line 34: It's so cool that with Postgres enums you get db level constraint-like validations for free

PR #126, opened 2020-04-16
Remove metrics definition from services

hvilloria:
## What does this PR do?

since the new idea is to generate metrics each day, some models and services like time intervals and metrics definitions were removed. Review turnaround flow has changed so the test use cases were changed too.

The main purpose is to process metrics at the end of every day, in what way we will be able to process week metrics, monthly metrics, etc. from these daily metrics.

issue: #120


app/jobs/process_metrics_job.rb
```diff
@@ -7,6 +7,6 @@ class ProcessMetricsJob < ApplicationJob
   ##
   # Delegates the processing to the MetricsDefinitionProcessor service.
   def perform
-    Processors::MetricsDefinition.call
+    Processors::ReviewTurnaround.call
```
hdamico, line 10: Shouldn't we change the file name also? For example `process_review_turnaround_job`

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -1,34 +1,43 @@
 module Metrics
   module ReviewTurnaround
-    class PerUserProject < Processors::Metric
+    class PerUserProject < BaseService
+      BATCH_SIZE = 500
+
+      def call
+        process
```
horacio, line 7: Why don't we have `process` body here?

app/services/processors/review_turnaround.rb
```diff
@@ -0,0 +1,17 @@
+module Processors
+  class ReviewTurnaround < BaseService
+    ENTITIES = ['UserProject'].freeze
+
+    def call
+      process
```
horacio, line 6: Same comment posted in `app/services/metrics/review_turnaround/per_user_project.rb`. Is this some convention agreed upon?

hdamico: Is a way to encapsulate the `process` method

hvilloria: exactly @horacio, others processors and have it like that so i wanted to keep the consistency about how are made the services. as you commented it is like a convention

horacio: Cool!

horacio:
:shipit:

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -1,34 +1,43 @@
...
         user.users_projects.detect { |user_project| user_project.project_id == project.id }
       end
 
       def calculate_turnaround(review)
-        review.review_request.created_at.to_i - review.opened_at.to_i
+        review.opened_at.to_i - review.review_request.created_at.to_i
+      end
+
+      def create_metric(entity, turnaround)
+        Metric.create!(ownable: entity, value: turnaround, name: :review_turnaround)
```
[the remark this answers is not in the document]
hvilloria: * what would be the timestamp for?

* i think it is not possible because we pretend to process every events at some point during the day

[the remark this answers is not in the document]
santiagovidal: This metrics tells me which entity it applies to, the value, the name of the metric, but I have no idea what time period it applies to.

PR #129, opened 2020-04-21
Events reprocessor script

hdamico:
## What does this PR do?

Generates a script to reprocess all the events that we have in production before all the models changes.

Resolves #XX

### DIFF
--- a/lib/events_processor.rb
+++ b/lib/events_processor.rb
@@ -7,8 +7,9 @@ class EventsProcessor
       errors = []
 
       retrieve_reviews do |event|
-        Projects::Builder.call(event.data['repository'])
-        event_name = event.data['event']
+        payload = event.data
+        Projects::Builder.call(payload['repository'])
+        event_name = payload['event']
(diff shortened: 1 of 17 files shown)


lib/events_processor.rb
```diff
@@ -0,0 +1,143 @@
+# The intention of this script is to reprocess all the events stored in Events table all over again.
```
santiagovidal, line 1: My concern with this script is that it's basically duplicating a lot of the logic we have in the webhooks.

I think it's fine if we use this script this time to process the events, but if the idea is to have a script that we can use whenever we want, we need a different approach.

hdamico: Not really, the builders do a `find_or_initialize` different from the other builder which does a `find_or_create` this is because we want to update also the ones that we already have. I tried to use the most, the methods that we already have.
I can recycle some methods, for example `find_or_create_user`, but I would have to inherit from `BaseService` which I don't believe it is ok.

PR #137, opened 2020-04-23
Hotfix

hdamico:
## What does this PR do?

Parse event name

### DIFF
--- a/app/services/github_service.rb
+++ b/app/services/github_service.rb
@@ -1,7 +1,7 @@
 class GithubService < BaseService
   def initialize(payload:, event:)
     @payload = payload
-    @event = event
+    @event = resolve_event_name(event)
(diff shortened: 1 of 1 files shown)


app/services/github_service.rb
```diff
@@ -28,6 +28,12 @@ def handle_event
     end
   end
 
+  def resolve_event_name(event)
```
santiagovidal, line 31: Can you add a brief comment explaining why this is necessary?

hdamico: Yep! Sorry, as a hotfix I mean to do it quick. We are receiving `pull_request_review` instead of `review` and `pull_request_review_comment` instead of `review_comment`. So in this way we make sure to get the appropriate event name

PR #142, opened 2020-04-23
One metric per user

hdamico:
## What does this PR do?

Fixes the case when a user makes reviews in different pull requests. Before this, we were creating two metric for the same user in the same day, now we ask if a metric already exists and if it is, we add the review turnaround time to the metric value.


app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -36,8 +36,12 @@ def calculate_turnaround(review)
         review.opened_at.to_i - review.review_request.created_at.to_i
       end
 
-      def create_metric(entity, turnaround)
-        Metric.create!(ownable: entity, value: turnaround, name: :review_turnaround)
+      def create_or_update_metric(entity, turnaround)
+        metric = Metric.find_or_initialize_by(ownable: entity,
```
santiagovidal, line 40: Shouldn't we find by the name of the metric, too?

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -36,8 +36,12 @@ def calculate_turnaround(review)
...
+                                              created_at: Time.zone.today.all_day)
+        return metric.update!(value: ((turnaround + metric.value) / 2)) if metric.persisted?
+
+        metric.update!(value: turnaround, name: :review_turnaround)
```
santiagovidal, line 44: Same comment here, we should update just the value

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -2,6 +2,7 @@ module Metrics
   module ReviewTurnaround
     class PerUserProject < BaseService
       BATCH_SIZE = 500
+      INTERVAL = Time.zone.today.all_day
```
santiagovidal, line 5: I think we should use a method for this. If I remember correctly, Rails will cache this value and it will only change it if you reset the server.

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -10,22 +11,42 @@ def call
...
 
-            create_metric(entity, turnaround)
+            create_or_update_metric(entity, turnaround)
           end
         end
+        calculate_avg
       end
 
       def filtered_reviews_ids
         Events::Review.joins(:review_request, owner: :users_projects, pull_request: :project)
```
santiagovidal, line 26: I'm wondering if we are using these join statements, given that we just select the ids.

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -10,22 +11,42 @@ def call
       private
 
       def process
-        filtered_reviews_ids.each_slice(BATCH_SIZE) do |batch|
-          Events::Review.eager_load(:review_request).find(batch).lazy.each do |review|
+        filtered_reviews_ids.lazy.each_slice(BATCH_SIZE) do |batch|
+          Events::Review.eager_load(:review_request).find(batch).each do |review|
```
santiagovidal, line 15: The JOIN statements should probably go here.

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -36,8 +57,12 @@ def calculate_turnaround(review)
         review.opened_at.to_i - review.review_request.created_at.to_i
       end
 
-      def create_metric(entity, turnaround)
-        Metric.create!(ownable: entity, value: turnaround, name: :review_turnaround)
+      def create_or_update_metric(entity, turnaround)
+        metric = Metric.find_or_initialize_by(ownable: entity,
+                                              created_at: Time.zone.today.all_day)
```
santiagovidal, line 62: We have to add `name: 'review_turnaround'` to the query, otherwise we could be updating a different metric from the same entity.

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -10,22 +11,42 @@ def call
...
+        filtered_reviews_ids.lazy.each_slice(BATCH_SIZE) do |batch|
+          Events::Review.eager_load(:review_request).find(batch).each do |review|
             entity = find_user_project(review.owner, review.pull_request.project)
             turnaround = calculate_turnaround(review)
 
-            create_metric(entity, turnaround)
+            create_or_update_metric(entity, turnaround)
           end
         end
+        calculate_avg
```
santiagovidal, line 22: I don't think this solves the problem we discussed. What if a user reviews several PRs for the same project in the same day? If that happens, you would call `create_or_update_metric(entity, turnaround)` several times for the same user-project pair (one for each review), and you would only keep the value that was processed last.

If you want we can hop on a call with @hvilloria and find a solution

app/services/metrics/review_turnaround/per_user_project.rb
hdamico: Yes! You are right! Thanks to let me know that

app/services/metrics/review_turnaround/per_user_project.rb
hdamico: Those joins are just to filter the reviews, like a condition for a Review to be included in the result. Do you think deleting this won't change the result?

app/services/metrics/review_turnaround/per_user_project.rb
hdamico: Yep, you are right! I'm using the relations without the joins and eager_loads

app/services/metrics/review_turnaround/per_user_project.rb
hdamico: Yes! I forgot

.rubocop.yml
```diff
@@ -28,6 +28,8 @@ Metrics/AbcSize:
   # The ABC size is a calculated magnitude, so this number can be an Integer or
   # a Float.
   Max: 15
+  Exclude:
+    - lib/events_processor.rb
```
santib, line 32: It'd be good to not exclude it 😇

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -3,29 +3,57 @@ module ReviewTurnaround
...
 
       def process
-        filtered_reviews_ids.each_slice(BATCH_SIZE) do |batch|
-          Events::Review.eager_load(:review_request).find(batch).lazy.each do |review|
-            entity = find_user_project(review.owner, review.pull_request.project)
-            turnaround = calculate_turnaround(review)
+        ActiveRecord::Base.transaction do
+          filtered_reviews_ids.lazy.each_slice(BATCH_SIZE) do |batch|
+            Events::Review.joins(:project, :review_request, :pull_request, owner: :users_projects)
+                          .eager_load(:project, :review_request, owner: :users_projects)
```
santib, line 20: will this perform good enough when having a lot of data?

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -3,29 +3,57 @@ module ReviewTurnaround
...
       end
 
       private
 
       def process
-        filtered_reviews_ids.each_slice(BATCH_SIZE) do |batch|
-          Events::Review.eager_load(:review_request).find(batch).lazy.each do |review|
-            entity = find_user_project(review.owner, review.pull_request.project)
-            turnaround = calculate_turnaround(review)
+        ActiveRecord::Base.transaction do
```
santib, line 17: should the transaction be here or inside the loop? Asking because if it fails maybe we can keep the iterations that were processed correctly

app/services/metrics/review_turnaround/per_user_project.rb
hdamico: Mmm I found it necessary because `joins` is being used as filter (INNER JOIN), and then `eager_load` to load in memory the necessary relations. Is there another way to make this?

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -36,8 +64,18 @@ def calculate_turnaround(review)
...
 
-      def create_metric(entity, turnaround)
-        Metric.create!(ownable: entity, value: turnaround, name: :review_turnaround)
+      def create_or_update_metric(entity, turnaround)
+        metric = Metric.find_or_initialize_by(ownable: entity,
+                                              created_at: Time.zone.today.all_day,
+                                              name: :review_turnaround)
+        return metric.update!(value: (turnaround + metric.value)) if metric.persisted?
+
+        metric.update!(value: turnaround)
```
santib, line 73: you are doing this because `metric.value` can be `nil` ? If yes:
maybe we can make that column not nullable
or
you can do `metric.update!(value: (turnaround + metric.value.to_i))` because `nil.to_i == 0`

santib: or `to_f` instead of `to_i` depending on the data type

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -36,8 +64,18 @@ def calculate_turnaround(review)
...
+      end
+
+      def metric_interval
+        @interval unless @interval.nil?
+        @metric_interval ||= Time.zone.today.all_day
```
santib, line 78:
```suggestion
        @metric_interval ||= @interval || Time.zone.today.all_day
```

app/services/metrics/review_turnaround/per_user_project.rb
hdamico: I need to keep `calculate_avg` inside the transaction because if something fails inside `calculate_avg` I would have Metrics with wrong data in the database, because I don't have the correct value of the Metric until `calculate_avg` happens.

app/services/metrics/review_turnaround/per_user_project.rb
hdamico: No, I made it in that way just to skip this:
```
metric.value = turnaround
metric.save!
```

app/services/metrics/review_turnaround/per_user_project.rb
santib: ok, let's go ahead and we can improve memory usage/performance later if needed

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -3,29 +3,57 @@ module ReviewTurnaround
...
-                      .where(opened_at: Time.zone.today.all_day)
+        Events::Review.joins(:project, :review_request, :pull_request, owner: :users_projects)
+                      .where(opened_at: metric_interval)
                       .order(:pull_request_id, :opened_at)
-                      .map(&:id)
+                      .pluck(Arel.sql('DISTINCT ON (reviews.pull_request_id) reviews.id'))
+      end
+
+      def pull_requests_count_per_user_project
+        Events::Review.joins(:review_request, :owner, :pull_request, :project)
```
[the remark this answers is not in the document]
santiagovidal: Actually.. we probably need to join with `review_request` to make sure that we count reviews that were requested, but do we need to join with the other tables?

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -3,29 +3,52 @@ module ReviewTurnaround
...
           end
+          calculate_avg(entities)
         end
       end
 
       def filtered_reviews_ids
-        Events::Review.joins(:review_request, owner: :users_projects, pull_request: :project)
-                      .select('DISTINCT ON (reviews.pull_request_id) reviews.id')
-                      .where(opened_at: Time.zone.today.all_day)
+        Events::Review.where(opened_at: metric_interval)
```
santiagovidal, line 34: Do we need to make sure we only count reviews that were requested?

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -3,29 +3,52 @@ module ReviewTurnaround
...
                       .order(:pull_request_id, :opened_at)
-                      .map(&:id)
+                      .pluck(Arel.sql('DISTINCT ON (reviews.pull_request_id) reviews.id'))
+      end
+
+      def pull_requests_count_per_user_project(entities)
+        entities_count = Hash.new(0)
+        entities.map { |value| [entities_count[value] += 1, value] }
```
santiagovidal, line 41: I feel like the `value` variable should be `entity` in this `map`. It was hard for me to understand until I realized that :)

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -3,29 +3,52 @@ module ReviewTurnaround
...
+                .reject { |value| value.first == 1 }
```
santiagovidal, line 42: I think you can write this as:

```
.reject { |count, _entity| count == 1 }
```

app/services/metrics/review_turnaround/per_user_project.rb
```diff
@@ -3,29 +3,53 @@ module ReviewTurnaround
...
+                      .where(opened_at: metric_interval)
                       .order(:pull_request_id, :opened_at)
-                      .map(&:id)
+                      .pluck(Arel.sql('DISTINCT ON (reviews.pull_request_id) reviews.id'))
+      end
+
+      def pull_requests_count_per_user_project(entities)
+        entities_count = Hash.new(0)
+        entities.map { |entity| [entities_count[entity] += 1, entity] }
+                .reject { |count, _entity| count == 1 }
```
santib, line 43: I don't get this logic, what do we want to get? something like this?
`[[1, userproject1], [1, userproject2], [2, userproject1], [3, userproject1]]` 
and after reject 
`[[2, userproject1], [3, userproject1]]`?

shouldn't we just have:
```
{
  userproject1: 3,
  userproject2: 1
}
```
?

hdamico: Fixed!

PR #144, opened 2020-04-28
renders review turnaround process

hvilloria:
## What does this PR do?

* renders review turnaround metric for a given project.


app/services/queries/daily_metrics.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+    end
+
+    def users_project
+      @users_project ||= Project.find(@project_id).users_projects
+    end
+
+    def format_data(metrics)
+      dates_and_values = Hash.new
+      metrics.each do |metric|
```
hdamico, line 24: What about an `inject` here?
For example, `metrics.each.inject({})`.

app/views/projects/review_turnaround.erb
```diff
@@ -0,0 +1,3 @@
+<div class='p-3'>
+  <%= line_chart  @metrics, ytitle: "Time as minutes", suffix: " minutes" %>
```
santiagovidal, line 2: Can we make it to hours?

db/seeds.rb
```diff
@@ -8,3 +8,17 @@
 if Rails.env.development?
   AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password')
 end
+
+project = FactoryBot.create(:project, name: 'rs-code-review-metrics')
```
santiagovidal, line 12: We should run all this only in `development`.

app/services/queries/daily_metrics.rb
hvilloria: much better!

horacio: Other option: `metrics.each_with_object({}) { |metric, hash| ... }`

app/views/projects/review_turnaround.erb
santiagovidal: For example if a review took 30 minutes, the graph should show 0.5 hours.

hvilloria: got it! i will review it

santiagovidal: You didn't change this 🤔

hvilloria: sorry, miss click

PR #148, opened 2020-04-29
Hotfix/script param

hdamico:
## What does this PR do?

Fixes script

### DIFF
--- a/lib/events_processor.rb
+++ b/lib/events_processor.rb
@@ -3,10 +3,10 @@
 
 class EventsProcessor
   class << self
-    def process
+    def process(entity)
       errors = []
 
-      retrieve_reviews do |event|
+      retrieve_entity_records(entity) do |event|
         payload = event.data
         event_name = resolve_event_name(event.name)
         Projects::Builder.call(payload['repository'])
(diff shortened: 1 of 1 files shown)


lib/events_processor.rb
```diff
@@ -17,8 +17,9 @@ def process
       Rails.logger.error errors unless errors.empty?
     end
 
-    def retrieve_reviews
-      Event.where('data ?| array[:keys]', keys: Event::TYPES).find_each.lazy.each do |event|
+    def retrieve_entity_records(entity)
+      entity = Event::TYPES if entity == 'all'
+      Event.where('data ?| array[:keys]', keys: entity).find_each.lazy.each do |event|
```
horacio, line 22: What do you think of...

```suggestion
    def retrieve_entity_records(types)
      types = Event::TYPES if types == 'all'
      Event.where('data ?| array[:keys]', keys: types).find_each.lazy.each do |event|
```

lib/events_processor.rb
```diff
@@ -17,8 +17,9 @@ def process
       Rails.logger.error errors unless errors.empty?
     end
 
-    def retrieve_reviews
-      Event.where('data ?| array[:keys]', keys: Event::TYPES).find_each.lazy.each do |event|
+    def retrieve_entity_records(types)
+      types = Event::TYPES if types == 'all'
```
santiagovidal, line 21: Perhaps we can do something like `types ||= Event::TYPES` and `def process(entity = nil)` so that we don't have to pass a string (`all`)

PR #150, opened 2020-04-30
Cron job

hdamico:
## What does this PR do?

Creates a cron job to execute review turnaround metrics generator job every day at 23:30.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ gem 'jquery-rails', '~> 4.3', '>= 4.3.5'
 
 gem 'chartkick', '~> 3.3', '>= 3.3.1'
 
-gem 'clockwork', '~> 1.2'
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -48,6 +48,8 @@ gem 'jquery-rails', '~> 4.3', '>= 4.3.5'
 
 gem 'chartkick', '~> 3.3', '>= 3.3.1'
 
+gem 'clockwork', '~> 1.2'
```
santib, line 51: I'm not really a fan of clockwork, I think these could be better alternatives: 
https://github.com/ondrejbartas/sidekiq-cron
https://github.com/moove-it/sidekiq-scheduler

wdyt?

hdamico: Great! I don't need a new worker on Heroku. Thanks for the approach

PR #155, opened 2020-05-05
Backfill blog posts

SandroDamilano:
## What does this PR do?
In the objective to be able to successfully import all blog posts from Wordpress, this PR adds:
- A `WordpressService`, in charge of accessing the Wordpress API
- A `BlogPostCategorizer`, in charge of assigning a specific technology to the blog post
- A `BlogPostsUpdater`, in charge of taking the posts on Wordpress and storing them in the DB
- A `BlogPostsUpdaterJob`, in charge of kicking off the posts updating process


Gemfile
```diff
@@ -48,6 +48,10 @@ gem 'jquery-rails', '~> 4.3', '>= 4.3.5'
 
 gem 'chartkick', '~> 3.3', '>= 3.3.1'
 
+gem 'stub_env', '~> 1.0'
+
+gem 'webmock', '~> 3.8'
+
```
santib, line 54: I think these gems should be added within the `test` group

Gemfile.lock
```diff
@@ -1,43 +1,43 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    actioncable (6.0.2)
-      actionpack (= 6.0.2)
+    actioncable (6.0.2.2)
+      actionpack (= 6.0.2.2)
```
santib, line 5: why did Rails got upgraded?

Gemfile
SandroDamilano: My bad, you're right!

app/services/blog_post_categorizer.rb
```diff
@@ -0,0 +1,17 @@
+class BlogPostCategorizer
+  def technology_for(blog_post_payload)
+    Technology.all.detect(default_technology) do |technology|
```
santib, line 3: why not something like:
```suggestion
    Technology.other.map do |technology|
```

looks simpler to me

app/services/wordpress_service.rb
```diff
@@ -0,0 +1,61 @@
...
+    posts += response_body[:posts]
+
+    if response_body[:meta]
+      blog_posts(posts: posts, next_page_token: response_body[:meta][:next_page])
+    else
+      posts
+    end
+  end
+
+  def access_token
```
santib, line 19: maybe all these methods below can be private?

app/jobs/blog_posts_updater_job.rb
```diff
@@ -0,0 +1,7 @@
+class BlogPostsUpdaterJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    Processors::BlogPostsUpdater.new.call
```
hdamico, line 5: As convention we use `BaseService` to make `Processors::BlogPostsUpdater.call` instead of using the `new`.

app/services/processors/blog_posts_updater.rb
```diff
@@ -0,0 +1,25 @@
...
+  class BlogPostsUpdater
+    def call
+      wordpress_service.blog_posts.each do |blog_post|
+        BlogPost.create_with(
+          published_at: blog_post[:date],
+          slug: blog_post[:slug],
+          status: blog_post[:status],
+          url: blog_post[:URL],
+          technology: categorizer.technology_for(blog_post)
+        ).find_or_create_by!(blog_id: blog_post[:ID])
```
hdamico, line 11: What about using only `find_or_create_by` with a block?
```
BlogPost.find_or_create_by!(blog_id: blog_post[:ID]) do |blog_post|
  blog_post.published_at = blog_post[:date],
  blog_post.slug = blog_post[:slug],
  blog_post.status = blog_post[:status],
  blog_post. url = blog_post[:URL],
  blog_post.technology = categorizer.technology_for(blog_post)
end
```
Just a suggestion, i don't really know which one is better.

app/services/blog_post_categorizer.rb
SandroDamilano: But is that actually equivalent? Doesn't `Technology.other` return a Technology object?
The code I wrote there means to look for the technology that matches any of the tags or categories. If none is found, then the default technology will be returned.

app/services/processors/blog_posts_updater.rb
SandroDamilano: I don't either, but I like yours best!

app/services/blog_post_categorizer.rb
santib: I thought `other`, was a scope 😂 sorry

SandroDamilano: Haha no worries! As much as I love the `detect` method, its expressiveness is far from being good 👎

PR #156, opened 2020-05-06
[Weekly Metrics]

hvilloria:
## What does this PR do?

* Implements Weekly metrics for users in a project.
* the main idea is to process the weekly metric every day after the daily metrics process so it can be updated as much as possible.


app/services/metrics/daily/review_turnaround/per_user_project.rb
```diff
@@ -0,0 +1,80 @@
+module Metrics
+  module Daily
```
hdamico, line 2: Being this daily, shouldn't the interval set to `Time.zone.now.all_day`? Or it could be any other day?

hvilloria: uhmm i think yes it could be, but i didn't change it thinking about the manually metric process.. like when we want to process an specific day

app/services/metrics/weekly/review_turnaround/per_user_project.rb
```diff
@@ -0,0 +1,52 @@
...
+            create_weekly(daily_metric)
+          end
+        end
+
+        def create_weekly(metric)
+          Metric.create!(
+            value: metric.value,
+            interval: :weekly,
+            name: :review_turnaround,
+            ownable: metric.ownable,
```
santiagovidal, line 29: I think it would be faster to just do:

```
ownable_type: metric.ownable_type,
ownable_id: metric.ownable_id
```

That way we don't have to query the db and load the object in memory. Maybe Rails already does this and doesn't do the query, I'm not sure.

app/services/metrics/weekly/review_turnaround/per_user_project.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+
+        private
+
+        def process
+          daily_metrics.find_each do |daily_metric|
+            weekly_metric = weekly_of(daily_metric.ownable)
+            if weekly_metric.present?
+              next weekly_metric.update!(
+                value: (daily_metric.value + weekly_metric.value) / current_time.wday
```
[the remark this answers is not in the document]
hvilloria: totally! i'm gonna change it!

app/services/metrics/weekly/review_turnaround/per_user_project.rb
```diff
@@ -0,0 +1,52 @@
...
+          process
+        end
+
+        private
+
+        def process
+          daily_metrics.find_each do |daily_metric|
+            weekly_metric = weekly_of(daily_metric.ownable)
+            if weekly_metric.present?
+              next weekly_metric.update!(
```
santib, line 15: code like this looks weird to me
```ruby
if something
  next update
end

create
```

don't you think it'd be nicer a create or update? Is this codebase using Rails 6? If yes, then we can use `upsert` https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert

spec/services/metrics/weekly/review_turaround/per_user_project_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    context 'generating a weekly metric for a user who has made his first review' do
+      let!(:metric) { create(:metric, ownable: users_project, value_timestamp: Time.zone.today) }
+      let(:weekly_metric_just_created) { Metric.find_by(interval: :weekly) }
+      it 'creates the metric with the same value the current daily metric' do
+        described_class.call
+        expect(metric.value).to eq(weekly_metric_just_created.value)
+      end
+
+      it 'creates just one weekly metric per daily metric' do
```
santib, line 16: one weekly metric per daily metric? shouldn't it be one weekly metric per seven daily metrics?

spec/services/metrics/weekly/review_turaround/per_user_project_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+               value: 60,
+               value_timestamp: current_time)
+      end
+
+      let(:current_number_day) { current_time.wday }
+
+      let!(:avg) { (weekly_metric.value + daily_metric.value) / current_number_day }
+
+      before do
+        allow(Time).to receive(:now) { current_time }
```
santib, line 44: you can freeze_time instead of mocking it https://api.rubyonrails.org/v5.2.4.2/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time

hvilloria: but it is possible to set a specific date? because i need that this date don't be weekend

hvilloria: i don't know if i should use freeze_time and travel in that case

spec/services/metrics/weekly/review_turaround/per_user_project_spec.rb
hvilloria: but in the context when the user has made his first review, in that case the value of the daily metric is the same as weekly metric and just one weekly metric is generated per daily metric generated in that day

santiagovidal:
@hvilloria we should close this right?

hvilloria:
> @hvilloria we should close this right?

yes we should

PR #159, opened 2020-05-07
Merge time metrics

hdamico:
## What does this PR do?

Processes and generates merge time metrics.


spec/services/metrics/merge_time/per_user_project_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+
+          let!(:third_pull_request) do
+            create(:pull_request,
+                   opened_at: Time.zone.now,
+                   merged_at: Time.zone.now + 2.hours,
+                   project_id: user_project.project_id,
+                   owner_id: user_project.user_id)
+          end
+
+          it 'creates just one metric' do
```
santiagovidal, line 64: Can we add another test to make sure the average is calculated correctly in this case?

spec/services/metrics/merge_time/per_user_project_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      it 'it generates the metric for the first project' do
+        expect(Metric.first.value.seconds).to eq(30.minutes)
+      end
+
+      it 'it generates the metric for the second project' do
+        expect(Metric.second.value.seconds).to eq(2.hours)
+      end
+    end
+
+    context 'when has a pull request in another project' do
```
santiagovidal, line 101: Should this be `when another user has a pull request in the same project`?

app/services/metrics/merge_time/per_user_project.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      def filtered_pull_requests
+        Events::PullRequest.includes(:project, owner: :users_projects)
+                           .where(opened_at: metric_interval)
+      end
+
+      def calculate_avg(entities)
+        entities.reject { |_entity, count| count == 1 }.each do |entity, count|
+          Metric.find_by!(ownable: entity, created_at: metric_interval).tap do |metric|
```
hvilloria, line 37: we should use the value_timestamp attr instead created_at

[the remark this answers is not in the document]
hdamico: God! I always miss that, thanks

spec/services/metrics/merge_time/per_user_project_spec.rb
hdamico: Is the same user in different project. In the `third_pull_request` is where i changed the project ID and not the user ID.

app/services/metrics/merge_time/per_user_project.rb
hdamico: And in the `create_or_update_metric` too right?

app/services/base_metric_service.rb
```diff
@@ -0,0 +1,21 @@
+class BaseMetricService < BaseService
+  def calculate_avg(entities)
+    entities.reject { |_entity, count| count == 1 }.each do |entity, count|
+      Metric.find_by!(ownable: entity, value_timestamp: metric_interval, name: :merge_time)
```
santiagovidal, line 4: The `name` should be an argument to the method

app/services/base_metric_service.rb
```diff
@@ -0,0 +1,21 @@
...
+        metric.value = metric.value / count
+        metric.save!
+      end
+    end
+  end
+
+  def create_or_update_metric(entity, merge_time)
+    metric = Metric.find_or_initialize_by(ownable: entity,
+                                          value_timestamp: Time.zone.today.all_day,
+                                          name: :merge_time)
```
santiagovidal, line 15: Same here

app/services/metrics/merge_time/per_user_project.rb
```diff
@@ -0,0 +1,48 @@
...
+          calculate_avg(entities)
+        end
+      end
+
+      def filtered_pull_requests
+        Events::PullRequest.includes(:project, owner: :users_projects)
+                           .where(merged_at: metric_interval)
+      end
+
+      def find_user_project(user, project)
```
santiagovidal, line 35: We can move this method to the top class, right?

app/services/base_metric_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    entities.reject { |_entity, count| count == 1 }.each do |entity, count|
+      Metric.find_by!(ownable: entity, value_timestamp: metric_interval, name: :merge_time)
+            .tap do |metric|
+        metric.value = metric.value / count
+        metric.save!
+      end
+    end
+  end
+
+  def create_or_update_metric(entity, merge_time)
```
santiagovidal, line 12: `merge_time` should be `metric_value` or something generic like that

db/migrate/20200506182951_add_value_to_metric_name_enum.rb
```diff
@@ -0,0 +1,16 @@
...
+      ALTER TYPE metric_name ADD VALUE 'merge_time';
+    SQL
+  end
+
+  def down
+    execute <<-SQL
+      ALTER TYPE metric_name RENAME TO metric_name_old;
+      CREATE TYPE metric_name AS ENUM ('review_turnaround', 'blog_visits');
+      ALTER TABLE metrics ALTER COLUMN name TYPE metric_name USING name::text::metric_name;
+      DROP TYPE metric_name_old;
```
santib, line 13: this SQL doesn't look like the inverse of the `up` method. what am I missing?

hdamico: You can't just delete an item from an ENUM, but you can add. So the only way to "delete" from an ENUM is changing the name of the current ENUM and create a new one with the old values.

app/jobs/process_merge_time_metrics_job.rb
```diff
@@ -0,0 +1,7 @@
+class ProcessMergeTimeMetricsJob < ApplicationJob
```
hvilloria, line 1: this will be a cron?

hdamico: Ohhh yes! I will put it in the scheduler

hvilloria: should we assign them different names? something like `queue_as :merge_time` so it can be readable in sidekiq,  what do you think?

PR #162, opened 2020-05-08
Fix requested reviewers creation flow

hdamico:
## What does this PR do?

Fixes how requested reviewers are created. Now GitHub sends you all the reviewers at once and also, if there are CODEOWNERS repeats the payload on every code owner. So, this PR adds a validation in `ReviewRequest` where it can't be more than one `ReviewRequest` with the same `pull_request_id`, `owner_id` and `reviewer_id`, and also at the creation (`ActionHandlers::PullRequest#review_requested`) an iteration of every reviewer in the payload.

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -110,7 +110,10 @@ detectors:
     accept: []
   UncommunicativeVariableName:
     enabled: true
-    exclude: ['EventsProcessor#process']
+    exclude: [
+      'EventsProcessor#process',
+      'ActionHandlers::PullRequest#review_requested'
+    ]
     reject:
       - "/^.$/"
       - "/[0-9]$/"
--- a/app/models/review_request.rb
+++ b/app/models/review_request.rb
@@ -24,7 +24,7 @@
 #  fk_rails_...  (reviewer_id => users.id)
 #
 class ReviewRequest < ApplicationRecord
-  enum state: { active: 'active', removed: 'removed' }
+  enum state: { active: 'active', removed: 'removed', reviewed: 'reviewed' }
 
   belongs_to :owner, class_name: 'User',
                      foreign_key: :owner_id,
@@ -39,5 +39,5 @@ class ReviewRequest < ApplicationRecord
                      dependent: :destroy
 
   validates :state, inclusion: { in: states.keys }
-  validates :pull_request_id, uniqueness: { scope: %i[owner_id reviewer_id] }
+  validates :pull_request_id, uniqueness: { scope: %i[owner_id reviewer_id state] }
 end
--- a/app/models/review_request.rb
+++ b/app/models/review_request.rb
@@ -39,4 +39,5 @@ class ReviewRequest < ApplicationRecord
                      dependent: :destroy
 
   validates :state, inclusion: { in: states.keys }
+  validates :pull_request_id, uniqueness: { scope: %i[owner_id reviewer_id] }
 end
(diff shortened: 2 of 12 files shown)


app/models/review_request.rb
```diff
@@ -39,4 +39,5 @@ class ReviewRequest < ApplicationRecord
                      dependent: :destroy
 
   validates :state, inclusion: { in: states.keys }
+  validates :pull_request_id, uniqueness: { scope: %i[owner_id reviewer_id] }
```
hvilloria, line 42: 👍

santiagovidal: Does this mean we won't save review requests when the owner of the PR does requested changes and requests a new review?

hdamico: Current_ly we only need the first one right? Or should we save everyone?

app/services/action_handlers/pull_request.rb
```diff
@@ -34,8 +34,12 @@ def review_request_removed
...
       pr_data = @payload['pull_request']
       owner = find_or_create_user(pr_data['user'])
-      reviewer = find_or_create_user(pr_data['requested_reviewers'].first)
-      @entity.review_requests.create!(owner: owner, reviewer: reviewer)
+      pr_data['requested_reviewers'].each do |raw_reviewer|
+        reviewer = find_or_create_user(raw_reviewer)
+        @entity.review_requests.create!(owner: owner, reviewer: reviewer)
+      end
+    rescue ActiveRecord::RecordInvalid => e
+      raise unless e.message == 'Validation failed: Pull request has already been taken'
```
santib, line 42: Some questions:
1. Should the rescue be inside of the `pr_data['requested_reviewers'].each` block? So that it continues with the rest of the elements?
2. If it already exists, don't you want to update it? If yes, you can use Rails 6 new method `create_or_find_by` https://blog.bigbinary.com/2019/03/25/rails-6-adds-create_or_find_by.html

hdamico: `create_or_find_by` won't cost more memory by finding the record? And for the first question, yes it could go inside the block.

santib: maybe.. if you don't need to update the object if already created (do you?) then you don't need to use `create_or_find_by`

hdamico: Do you what? I will update it, 'cause I realize that the reviewer could be removed and the state would be 'removed', so I will update the state to 'active'.

.reek.yml
```diff
@@ -110,7 +110,10 @@ detectors:
     accept: []
   UncommunicativeVariableName:
     enabled: true
-    exclude: ['EventsProcessor#process']
```
hdamico, line 113: Test

app/services/action_handlers/pull_request.rb
```diff
@@ -26,16 +26,30 @@ def closed
...
+      rescue ActiveRecord::RecordInvalid => e
+        raise unless e.message == 'Validation failed: Pull request has already been taken'
+      end
+    end
+
+    def remove_review_requests(owner, reviewers)
+      @entity.review_requests
+             .includes(:owner)
+             .where.not(reviewer: User.where(github_id: reviewers))
+             .where(owner: owner)
```
santiagovidal, line 51: Why do we have to check for the `owner` here? `@entity` is the PR, right? That means we already know that the review requests belong to the PR, and therefore the owner.

PR #163, opened 2020-05-11
[Weekly-metrics-v2]

hvilloria:
## What does this PR do?

* process all daily metrics at the end of the day to create a weekly metric, this metric is being updated if a new daily metric for the same type is created.
* process all daily metrics independently of the entity and the type (name) of metric.

### DIFF
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -25,4 +25,8 @@ class Metric < ApplicationRecord
 
   validates :interval, inclusion: { in: intervals.keys }
   validates :name, inclusion: { in: names.keys }
+
+  scope :today_daily_metrics, lambda {
+    where(value_timestamp: Time.zone.today.all_day, interval: :daily)
+  }
 end
--- a/app/services/metrics/weekly/all_metrics.rb
+++ b/app/services/metrics/weekly/all_metrics.rb
@@ -1,0 +1,50 @@
+module Metrics
+  module Weekly
+    class AllMetrics < BaseService
+      def call
+        process
+      end
+
+      private
+
+      def process
+        Metric.today_daily_metrics.find_each do |daily_metric|
+          create_or_update(daily_metric)
+        end
+      end
+
+      def create_or_update(daily_metric)
+        metric = Metric.find_or_initialize_by(
+          ownable: daily_metric.ownable,
+          value_timestamp: week_range,
+          interval: :weekly,
+          name: daily_metric.name
+        )
+
+        value = metric_avg(daily_metric)
+        return metric.update!(value: value) if metric.persisted?
+
+        metric.value = value
+        metric.value_timestamp = current_time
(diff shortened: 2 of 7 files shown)


app/services/metrics/weekly/all_metrics.rb
```diff
@@ -0,0 +1,50 @@
...
+        ).average(:value)
+      end
+
+      def week_range
+        @week_range ||= current_time.beginning_of_week..current_time.end_of_week
+      end
+
+      def current_time
+        @current_time = Time.zone.now
+      end
```
horacio, line 47: WDYT about just calling `Time.zone.now` on line 28? I don't think we can or should reuse this method.

hvilloria: the reason is because we are using `Time.zone.now` several times, the is to have precision every time it is used over the creation/querying the metrics, what do you think? 🤔  still prefer just the time zone? @horacio

app/models/metric.rb
```diff
@@ -25,4 +25,8 @@ class Metric < ApplicationRecord
 
   validates :interval, inclusion: { in: intervals.keys }
   validates :name, inclusion: { in: names.keys }
+
+  scope :today_daily_metrics, lambda {
```
hdamico, line 29: WDYT of using `->` instead of `lambda`?

hvilloria: at first i did it like that, but the analysis throw me a reek warning

hdamico: Alright! Did you find why is giving you that error?

hvilloria: sorry I got confused, was not a warning .. the warning was about the way i was doing the query for the ownable.. there is a lambda because the need of a block so i couldn't do something like `scope today_daily_metrics, -> {`
`code..`
`}`

i had to convert it into a lamba (Rubocop)

app/services/metrics/weekly/all_metrics.rb
horacio: Ahhh I see. I thought it was used just in one place. I think it's fine!

app/services/metrics/weekly/all_metrics.rb
```diff
@@ -0,0 +1,50 @@
...
+  module Weekly
+    class AllMetrics < BaseService
+      def call
+        process
+      end
+
+      private
+
+      def process
+        Metric.today_daily_metrics.find_each do |daily_metric|
```
santiagovidal, line 11: This method is `O(n2)` with the `Metrics` table, which could grow a lot. I think we can move forward with this solution though, because I can't think of a better one, but we're going to have to keep an eye on the performance of this job.

hvilloria: roger that

PR #165, opened 2020-05-13
Create blog post views metrics

SandroDamilano:
## What does this PR do?
Set up a processor (`BlogPostViewsUpdater`) to create visits metrics for a blog post

### DIFF
--- a/spec/services/processors/blog_post_views_updater_spec.rb
+++ b/spec/services/processors/blog_post_views_updater_spec.rb
@@ -55,18 +55,20 @@ describe Processors::BlogPostViewsUpdater do
           .to(current_month_views)
       end
 
-      context 'and last month as well' do
-        let(:publish_date) { Time.zone.now.last_month.beginning_of_month }
+      context 'and all months since it publication' do
+        let(:publish_date) { Time.zone.now.last_year }
 
         before do
-          create(
-            :metric,
-            name: Metric.names[:blog_visits],
-            interval: Metric.intervals[:monthly],
-            value: outdated_month_views,
-            value_timestamp: publish_date.next_month.end_of_month,
-            ownable: blog_post
-          )
+          (publish_date.to_date..Time.zone.today).map(&:end_of_month).uniq.each do |time|
+            create(
+              :metric,
+              name: Metric.names[:blog_visits],
+              interval: Metric.intervals[:monthly],
+              value: outdated_month_views,
+              value_timestamp: time.end_of_day,
+              ownable: blog_post
+            )
+          end
         end
 
         it 'does not access last month views as they will not have changed' do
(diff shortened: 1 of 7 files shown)


spec/services/processors/blog_post_views_updater_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+    let(:publish_date) { Time.zone.now.beginning_of_month }
+    let(:current_month_views) do
+      blog_post_views_payload['years'][publish_date.year.to_s]['months'][publish_date.month.to_s]
+    end
+    let(:api_service) { instance_double(WordpressService) }
+
+    before do
+      allow_any_instance_of(described_class).to receive(:wordpress_service).and_return(api_service)
+      allow(api_service)
```
santiagovidal, line 17: It's a not something that needs to be fixed now, but I usually prefer to use Webmock than to stub services. I'm interested in rest of the team's opinion on this as well.

spec/services/processors/blog_post_views_updater_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        .to receive(:blog_post_views)
+        .with(blog_post.blog_id)
+        .and_return(blog_post_views_payload)
+    end
+
+    it 'creates views metrics for the blog post' do
+      described_class.call(blog_post.blog_id)
+
+      created_metric = Metric.find_by(ownable: blog_post)
```
santiagovidal, line 26: I'd add the `name` condition to this query instead of an `expect`

Otherwise in the future when we have more metrics related to Blog posts, this test will fail

spec/services/processors/blog_post_views_updater_spec.rb
SandroDamilano: Yeah, I agree with that! I'll mock the call instead

spec/services/processors/blog_post_views_updater_spec.rb
SandroDamilano: I think it shouldn't if we keep the DB clean, but nevertheless I'd rather do what you said just in case

santiagovidal:
Let's mark that CodeClimate issue as invalid. Do we all agree?

SandroDamilano:
The CodeClimate issue about coverage should be fixed in the next PR (https://github.com/rootstrap/rs-code-review-metrics/pull/169)

spec/services/processors/blog_post_views_updater_spec.rb
SandroDamilano: I finally added it in a separate PR: https://github.com/rootstrap/rs-code-review-metrics/pull/169

PR #166, opened 2020-05-13
Rails performance and Honeybadger

hdamico:
## What does this PR do?

Implements Rails Performance for performance monitoring and HoneyBadger for exceptions monitoring.


Gemfile
```diff
@@ -50,6 +50,10 @@ gem 'chartkick', '~> 3.3', '>= 3.3.1'
 
 gem 'sidekiq-cron', '~> 1.1'
 
+gem 'rails_performance', '~> 0.9.1'
```
hvilloria, line 53: comment to test the github webhook

santib, line 53: can we also add rack mini profiler? I'd really like to learn about that tool 😬 https://github.com/MiniProfiler/rack-mini-profiler

config/initializers/rails_performance.rb
```diff
@@ -0,0 +1,8 @@
+if defined?(RailsPerformance)
+  RailsPerformance.setup do |config|
+    config.redis    = Redis.new(url: ENV.fetch('REDIS_PERFORMANCE_URL') { 'redis://localhost:6379/1' })
+    config.duration = 4.hours
+    config.enabled  = true
+    config.http_basic_authentication_enabled = false
```
santib, line 6: don't we need to have this enabled so that only admins access the performance panel?

hdamico: Sure, do i hardcode the credentials aswell as credentials for activeadmin? Or do you preffer to save them in the ENV and use the same for both?

Gemfile
hdamico: Sure!

config/initializers/rails_performance.rb
santib: maybe env variables is better, or if you can authorize by `AdminUser` then better

santib: not sure if it's possible with this gem

Gemfile
```diff
@@ -50,6 +50,10 @@ gem 'chartkick', '~> 3.3', '>= 3.3.1'
...
+
+gem 'honeybadger', '~> 4.0'
```
santib, line 55: should we also add exception hunter? or not yet? cc @brunvez

config/initializers/rails_performance.rb
hdamico: Ok, I will check, otherwise I use ENV

Gemfile
brunvez: We still need to release it to RubyGems, I plan to have it by Monday. I don't want to be a blocker so maybe we can do a different PR.

PR #167, opened 2020-05-14
Fix review request logic

hdamico:
## What does this PR do?

Fixes Review Request creation logic.
Now it will create a ReviewRequest unless there is a ReviewRequest with the same pull request, reviewer and the state active.
When a Review or a ReviewComment is submitted it will set the ReviewRequest to reviewed, so we can track if the ReviewRequest is already handled by the reviewer, if this happens and the owner of the pull request re request the reviewer to review if will create another ReviewRequest.
With this now we can track multiple reviews in the same pull request.

### DIFF
--- a/app/models/review_request.rb
+++ b/app/models/review_request.rb
@@ -24,7 +24,7 @@
 #  fk_rails_...  (reviewer_id => users.id)
 #
 class ReviewRequest < ApplicationRecord
-  enum state: { active: 'active', removed: 'removed', reviewed: 'reviewed' }
+  enum state: { active: 'active', removed: 'removed' }
 
   belongs_to :owner, class_name: 'User',
                      foreign_key: :owner_id,
@@ -39,5 +39,4 @@ class ReviewRequest < ApplicationRecord
                      dependent: :destroy
 
   validates :state, inclusion: { in: states.keys }
-  validates :pull_request_id, uniqueness: { scope: %i[owner_id reviewer_id state] }
 end
--- a/app/services/action_handlers/pull_request.rb
+++ b/app/services/action_handlers/pull_request.rb
@@ -29,9 +29,8 @@ module ActionHandlers
       pr_data = @payload['pull_request']
       removed_reviewer = find_or_create_user(@payload['requested_reviewer'])
       owner = find_or_create_user(pr_data['user'])
-      review_request = @entity.review_requests
-                              .find_by(owner: owner, reviewer: removed_reviewer, state: 'active')
-      review_request.removed! unless review_request.nil?
+      @entity.review_requests.find_by(owner: owner, reviewer: removed_reviewer, state: 'active')
+                             &.removed!
     end
 
     def review_requested
--- a/app/services/action_handlers/pull_request.rb
+++ b/app/services/action_handlers/pull_request.rb
@@ -26,10 +26,8 @@ module ActionHandlers
     end
 
     def review_request_removed
-      pr_data = @payload['pull_request']
       removed_reviewer = find_or_create_user(@payload['requested_reviewer'])
-      owner = find_or_create_user(pr_data['user'])
-      @entity.review_requests.find_by(owner: owner, reviewer: removed_reviewer, state: 'active')
+      @entity.review_requests.find_by(reviewer: removed_reviewer, state: 'active')
                              &.removed!
     end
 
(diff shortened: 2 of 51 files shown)


hdamico:
TODO

- [ ] Add condition in ReviewTurnaround where the review requests have to be 'reviewed'.

app/services/action_handlers/pull_request.rb
```diff
@@ -27,26 +27,20 @@ def closed
...
-
-    def remove_review_requests(reviewers)
-      @entity.review_requests
-             .includes(:owner)
-             .where.not(reviewer: User.where(github_id: reviewers))
-             .each(&:removed!)
+      review_request = @entity.review_requests.find_or_initialize_by(owner: owner,
+                                                                     reviewer: reviewer,
+                                                                     state: 'active')
+      review_request.save! unless review_request.persisted?
```
santiagovidal, line 43: Is it not the same to just do `find_or_create_by`?

app/services/action_handlers/review_comment.rb
```diff
@@ -20,5 +21,12 @@ def edited
     def deleted
       @entity.removed!
     end
+
+    def mark_review_request_as_reviewed
+      ReviewRequest.find_by(pull_request_id: @entity.pull_request_id,
```
santiagovidal, line 26: Can't you get the `review_request` directly with `@entity`?

app/services/action_handlers/pull_request.rb
```diff
@@ -26,27 +26,16 @@ def closed
...
-        raise unless e.message == 'Validation failed: Pull request has already been taken'
-      end
-    end
-
-    def remove_review_requests(reviewers)
-      @entity.review_requests
-             .includes(:owner)
-             .where.not(reviewer: User.where(github_id: reviewers))
-             .each(&:removed!)
+      @entity.review_requests.create(reviewer: reviewer, owner: owner, state: 'active').persisted?
```
santiagovidal, line 38: Why are we calling `persisted?`?

app/models/review_request.rb
```diff
@@ -24,7 +24,7 @@
 #  fk_rails_...  (reviewer_id => users.id)
 #
 class ReviewRequest < ApplicationRecord
-  enum state: { active: 'active', removed: 'removed', reviewed: 'reviewed' }
+  enum state: { active: 'active', removed: 'removed' }
```
santiagovidal, line 27: Why did we remove that state?

app/services/event_builder.rb
```diff
@@ -13,8 +13,10 @@ def find_pull_request
     Events::PullRequest.find_by!(github_id: @payload['pull_request']['id'])
   end
 
-  def find_first_review_request(pull_request, reviewer_id)
-    pull_request.review_requests.where(reviewer_id: reviewer_id).order(:created_at).first
+  def find_last_review_request(pull_request, reviewer_id)
+    pull_request.review_requests.where(reviewer_id: reviewer_id).last
+  rescue StandardError
```
santiagovidal, line 18: Can we be more specific here? What exception are we expecting if there is no review_request?

app/services/action_handlers/pull_request.rb
```diff
@@ -26,27 +26,16 @@ def closed
     end
 
     def review_request_removed
-      pr_data = @payload['pull_request']
-      reviewers_github_id = pr_data['requested_reviewers'].map { |reviewer| reviewer['id'] }
-      remove_review_requests(reviewers_github_id)
+      removed_reviewer = find_or_create_user(@payload['requested_reviewer'])
+      @entity.review_requests.find_by(reviewer: removed_reviewer, state: 'active')
```
santib, line 30: maybe the state wasn't active?

hdamico: We have active and removed. To check if it was reviewed just ask if it has al least one review.

app/models/review_request.rb
hdamico: I believe it is unecessary, we can check that by asking if that review request has at least one review.

app/services/action_handlers/pull_request.rb
hdamico: The method `create` by default returns the object, even if it no saved, so with `persisted?` we can be sure if the record was saved.

hdamico: Also, it does not hit the database twice.

santiagovidal: But do we use the result of the method?

hdamico: Rubocop finds using create without the exclamation mark a bad practice, so, i found this is an alternative for that, otherwise I would have to use a rescue to catch the exception. I guess that is a bad practice to don't check if the record was or not saved.

horacio: Sorry to reiterate on this issue, but could it be the case that it's the other way around?

hdamico: Yep, it says that you have to use the exclamation mark, but I searched that specific error and I found that you can do the same by adding the persisted? at the end

PR #168, opened 2020-05-15
Graph weekly metric

hvilloria:
## What does this PR do?

* Creates new service to query and build data for weekly metrics


app/services/queries/daily_metrics.rb
```diff
@@ -7,24 +7,23 @@ def initialize(project_id)
...
+      )
     end
 
     private
 
-    def users_project
-      @users_project ||= Project.find(@project_id).users_projects
+    def query
+      {
+        value_timestamp: (actual_time - FROM)..actual_time,
```
santiagovidal, line 20: Shouldn't we specify the `name` of the Metric?

app/views/shared/_sidebar.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="col container-height sidebar-border">
+  <div class="container-md">
+    <div class="row p-3">
+      <select
+        class="time-period-select js-states form-control sidebar-selector"
+        data-placeholder="periodo de tiempo"
```
santiagovidal, line 6: Spanish 😆

app/views/shared/_sidebar.html.erb
```diff
@@ -0,0 +1,14 @@
...
+      >
+        <option></option>
+        <option value="daily">Diario</option>
```
santiagovidal, line 9: Same here.

app/services/queries/weekly_metrics.rb
```diff
@@ -0,0 +1,34 @@
+module Queries
+  class WeeklyMetrics < BaseService
```
SandroDamilano, line 2: Could this and `DailyMetrics` inherit from another common superclass and put some shared code over there?

app/assets/javascripts/sidebar.js
```diff
@@ -0,0 +1,7 @@
+
+$(document).ready(function() {
+  $('.time-period-select').select2();
+  $( ".time-period-select" ).change(function(event) {
```
santib, line 4: this line is not consistent with the line above. are we using single our double quotes in JS? Also there are extra spaces around the selector

app/assets/javascripts/sidebar.js
```diff
@@ -0,0 +1,7 @@
...
+    window.location.replace(window.location.pathname + `?period=${event.target.value}`)
```
santib, line 5:
```suggestion
    window.location.replace(`${window.location.pathname}?period=${event.target.value}`)
```

santib, line 5: this seems a bit to manual, why don't you work with a form like here https://medium.com/swlh/advanced-filtering-for-your-rails-5-application-28c8da2d29b6 ?

app/views/shared/_sidebar.html.erb
hvilloria: 🤦‍♂️ my bad

app/services/queries/weekly_metrics.rb
hvilloria: i think both classes has some different important things like the constructor and the range of the time, but the way that they parse the data in fact was very similar for chartkick, so it was already moved to another class

app/assets/javascripts/sidebar.js
```diff
@@ -0,0 +1,12 @@
+
+$(document).ready(function() {
+  const optSelected = $('select.period-selection').children("option:selected").val();
+  if (optSelected === '') {
+    const url = window.location.href.split('?')[0]
+    window.history.replaceState({}, document.title, url);
+  }
```
hvilloria, line 7: @santib what do you think about this? the idea is remove the query string when user select 'empty' in dropdown

hvilloria:
@santib @santiagovidal please, check the simple form config and the form in sidebar 🙏 , to be sure that it is the best way

app/assets/javascripts/sidebar.js
santib: how would the query string link when selecting `empty` in dropdown without this change?

app/services/chartkick_data_builder.rb
```diff
@@ -0,0 +1,29 @@
...
+  def build_data(metrics)
+    metrics.each.inject({}) do |hash, metric|
+      hash.merge!(
+        metric.value_timestamp.strftime('%Y-%m-%d').to_s => value_in_hours_for(metric)
+      )
+    end
+  end
+
+  def value_in_hours_for(metric)
+    (metric.value.to_f / SECONDS).round(2)
```
santib, line 27: maybe you can use `1.hour.seconds` ?

app/models/users_project.rb
```diff
@@ -18,4 +18,6 @@ class UsersProject < ApplicationRecord
   has_many :metrics, as: :ownable, dependent: :destroy
 
   validates :user_id, uniqueness: { scope: :project_id }
+
+  delegate :name, to: :user
```
santib, line 22: can we prefix it? I mean, `name` could also be delegated to `project` and it would make sense 😅

app/assets/javascripts/sidebar.js
hvilloria: for daily metric it is like this: 
`http://localhost:3000/projects/20/metrics/review_turnaround?metric%5Bperiod%5D=daily`
withouth that change would be like this:
`http://localhost:3000/projects/20/metrics/review_turnaround?metric%5Bperiod%5D=`

app/services/queries/weekly_metrics.rb
SandroDamilano: Yeah, I agree that it was a smart move. I just felt like many of these methods where exactly the same (`call`, `users_project`, `current_time`) so we could keep them in just one place.
But as it's not much, if you think readability is better this way, I'll leave it up to you whether create a superclass or not 👌

app/assets/javascripts/sidebar.js
santib: ok I'm good leaving this 👌

app/controllers/projects_controller.rb
```diff
@@ -1,5 +1,21 @@
 class ProjectsController < ApplicationController
   def review_turnaround
-    @metrics = Queries::DailyMetrics.call(params[:project_id])
+    return if params.dig(:metric, :period).blank?
+
+    @metrics = if metric_params[:period] == 'weekly'
+                 Queries::WeeklyMetrics.call(project_id)
+               else
+                 Queries::DailyMetrics.call(project_id)
```
santib, line 8:
```suggestion
               elsif metric_params[:period] == 'daily'
                 Queries::DailyMetrics.call(project_id)
               else
                 raise UnknownDateRange
               end
```

minor comment, but maybe it's good to avoid confusion. e.g. if someone sends 'monthly' they'd still be getting daily values

app/services/queries/daily_metrics.rb
```diff
@@ -7,24 +7,24 @@ def initialize(project_id)
...
```
santib, line 20: I think instead of `(actual_time - FROM)` you could have `FROM.ago(actual_time)`

app/services/queries/daily_metrics.rb
```diff
@@ -7,24 +7,24 @@ def initialize(project_id)
...
+        interval: :daily,
+        name: :review_turnaround
```
santib, line 22: the file is called `app/services/queries/daily_metrics.rb` , the name is pretty generic, so I don't think `name: :review_turnaround` belongs here. Shouldn't it be passed as a param to the service?

app/views/shared/_sidebar.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <div class="row p-3">
+      <%= simple_form_for :metric, url: review_turnaround_project_metrics_path,
+        method: 'GET', html: { class: 'sidebar-form' } do |f| %>
+          <%= f.input :period,
+            input_html: {class: 'period-selection'},
+            required: false,
+            collection: [[],['Daily', 'daily'], ['Weekly', 'weekly']],
+            selected: value_from_query_param
+          %>
+      <% end %>
```
santib, line 12: correct me if I'm wrong but I don't think this belongs to the shared sidebar that'll be accessible in all of the views but only in some pages. I'm right then you might need to use something like
```suggestion
      <%= yield :custom_filters %>
```

and then in a view for `review turnaround`
```erb
  <% content_for :custom_filters do %>
    <%= simple_form_for :metric, url: review_turnaround_project_metrics_path,
        method: 'GET', html: { class: 'sidebar-form' } do |f| %>
          <%= f.input :period,
            input_html: {class: 'period-selection'},
            required: false,
            collection: [[],['Daily', 'daily'], ['Weekly', 'weekly']],
            selected: value_from_query_param
          %>
      <% end %>
  <% end %>
```

lib/templates/erb/scaffold/_form.html.erb
```diff
@@ -0,0 +1,14 @@
+<%%= simple_form_for(@<%= singular_table_name %>) do |f| %>
```
santib, line 1: what's this?

spec/models/user_spec.rb
```diff
@@ -46,4 +46,10 @@
     subject.login = nil
```
santib, line 46: would it be possible to add some `feature`/`system` tests using capybara and chrome?

https://guides.rubyonrails.org/testing.html#system-testing

app/services/queries/daily_metrics.rb
hvilloria: yes, @hdamico is adding it in his PR 👍

app/views/shared/_sidebar.html.erb
hvilloria: actually even the filter is going to be refactored, this was the initial idea to show the weekly metric but, my next card is to build the UI as we talked yesterday

lib/templates/erb/scaffold/_form.html.erb
hvilloria: the simple form installation added this file 🤔 , im going to remove it !

app/javascript/modules/graph/index.js
```diff
@@ -0,0 +1,16 @@
+const handleSidebarChange = () => {
+  let selectInput = document.querySelector('select.period-selection');
+  let optionSelected = selectInput.options[selectInput.selectedIndex].text;
+  const sidebarForm = document.querySelector('.sidebar-form');
+
+  if (optionSelected === '') {
+    const url = window.location.href.split('?')[0]
+    window.history.replaceState({}, document.title, url);
+  }
```
santib, line 9: for a future PR. I think this code and 
```js
  let selectInput = document.querySelector('select.period-selection');
  let optionSelected = selectInput.options[selectInput.selectedIndex].text;

  if (optionSelected === '') {
    const url = window.location.href.split('?')[0]
    window.history.replaceState({}, document.title, url);
  }
```
should be moved to a different file because it has nothing to do with handling the sidebar change, right? It's just something we want to run once the page is loaded

hvilloria: you are right, i will change it

PR #169, opened 2020-05-18
Create helper to stub Wordpress API requests

SandroDamilano:
## What does this PR do?
It adds the `WordpressApiMocker` to make stubbing the Wordpress API much easier when testing.

### DIFF
--- a/spec/support/helpers/wordpress_api_helper.rb
+++ b/spec/support/helpers/wordpress_api_helper.rb
@@ -1,0 +1,132 @@
+module WordpressApiHelper
+  def stub_blog_post_views_response(
+    blog_post_id,
+    blog_post_views_payload = create(:blog_post_views_payload)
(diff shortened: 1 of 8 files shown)


spec/support/helpers/wordpress_api_helper.rb
```diff
@@ -0,0 +1,132 @@
+module WordpressApiHelper
```
santib, line 1: I'd avoid the name `Helper` to avoid confusions with Rails Helpers that are intended to be used in views. WDYT about `WordpressMocker` ?

SandroDamilano: Sure! What about `WordpressAPIMocker`?

santib: sure

PR #174, opened 2020-05-22
Add processor to update blog visits metrics

SandroDamilano:
## What does this PR do?
Adds a processor called `BlogMetricsUpdater`, which first updates all the visits metrics of each blog post and then calculates those visits per technology.
As visits from previous months remain unchanged, it only iterates through the metrics of the current month (or since the month lastly updated).

### DIFF
--- a/app/services/processors/blog_metrics_updater.rb
+++ b/app/services/processors/blog_metrics_updater.rb
@@ -8,7 +8,7 @@ module Processors
     private
 
     def update_blog_post_visits_metrics
-      BlogPost.find_each(batch_size: 50) do |blog_post|
+      BlogPost.find_each(batch_size: 50).lazy.each do |blog_post|
         BlogPostViewsUpdater.call(blog_post.blog_id)
       end
     end
--- a/app/services/processors/blog_metrics_updater.rb
+++ b/app/services/processors/blog_metrics_updater.rb
@@ -1,0 +1,52 @@
+module Processors
+  class BlogMetricsUpdater < BaseService
+    def call
+      update_blog_post_visits_metrics
+      update_technologies_visits_metrics
+    end
+
+    private
+
+    def update_blog_post_visits_metrics
+      BlogPost.find_each(batch_size: 50) do |blog_post|
+        BlogPostViewsUpdater.call(blog_post.blog_id)
+      end
+    end
+
+    def update_technologies_visits_metrics
+      blog_views_by_technology_and_month.each do |pair_group, visits|
+        timestamp, technology_id = pair_group
+        update_technology_visits_metric(technology_id, visits, timestamp)
+      end
+    end
+
+    def blog_views_by_technology_and_month
+      Metric.where(name: Metric.names[:blog_visits], ownable_type: BlogPost.to_s)
+            .where('value_timestamp >= ?', latest_technology_metrics_updated_at)
+            .joins('JOIN blog_posts on metrics.ownable_id = blog_posts.id')
+            .group(:value_timestamp, :technology_id)
+            .sum(:value)
+    end
+
+    def latest_technology_metrics_updated_at
+      latest_visits_metrics_by_technology =
+        Metric.where(name: Metric.names[:blog_visits], ownable_type: Technology.to_s)
+              .group(:ownable_id)
+              .maximum(:value_timestamp)
+
+      latest_visits_metrics_by_technology.values.min || Time.zone.at(0)
+    end
+
+    def update_technology_visits_metric(technology_id, visits, timestamp)
+      metric = Metric.find_or_initialize_by(
+        ownable_id: technology_id,
+        ownable_type: Technology.to_s,
+        name: Metric.names[:blog_visits],
+        interval: Metric.intervals[:monthly],
+        value_timestamp: timestamp
+      )
+      metric.value = visits
+      metric.save!
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/processors/blog_metrics_updater.rb
```diff
@@ -0,0 +1,52 @@
...
+    def blog_views_by_technology_and_month
+      Metric.where(name: Metric.names[:blog_visits], ownable_type: BlogPost.to_s)
+            .where('value_timestamp >= ?', latest_technology_metrics_updated_at)
+            .joins('JOIN blog_posts on metrics.ownable_id = blog_posts.id')
+            .group(:value_timestamp, :technology_id)
+            .sum(:value)
+    end
+
+    def latest_technology_metrics_updated_at
+      latest_visits_metrics_by_technology =
```
santiagovidal, line 32: Can we memoize this? So that it doesn't run a query for each iteration.

spec/services/processors/blog_metrics_updater_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+        ownable: technology,
+        interval: Metric.intervals[:monthly],
+        name: Metric.names[:blog_visits]
+      )
+
+      expect(metric).to be_present
+    end
+  end
+
+  describe '#update_blog_post_visits_metrics' do
```
santiagovidal, line 38: I think testing private methods is not a good practice. Perhaps we can add these tests under the describe for `.call` and group them with contexts.

app/services/processors/blog_metrics_updater.rb
SandroDamilano: I could do it, absolutely, but I think it still runs just once.
It's used in `blog_views_by_technology_and_month`, and that method is called once to retrieve all the records. Then it just iterates on the results.
Or am I missing something?

spec/services/processors/blog_metrics_updater_spec.rb
SandroDamilano: I see what you say. I wanted to tackle this on a more unit-testing approach, so that it is better isolated.
If I were to move them to the `'.call'` describe, I would mess up the setup I'm creating in `'#update_technologies_visits_metrics'`, because the other method I run in `call` (`update_blog_post_visits_metrics`) would interfere on it (it creates the blog post visits metrics I use there to create the technology ones).

I think the alternative could be to move both and to create a separate processor for updating the technologies visits metrics (which may be the right move anyway), and test it in isolation in its own test suite.
What do you think?

app/services/processors/blog_metrics_updater.rb
santiagovidal: You're absolutely right, my bad.

spec/services/processors/blog_metrics_updater_spec.rb
santiagovidal: I don't think I fully understood, why would it interfere? If your blog post visits metrics are already present before you run `.call`, then `update_blog_post_visits_metrics` shouldn't do anything right?

SandroDamilano: Oh, you're right, it shouldn't! Thank you! I'll move them both, then

SandroDamilano: Both setups and assertions did still interfere with each other, so it wasn't that straightforward. So I ended up creating a separate processor for technologies, as I think I would end up doing it in the future anyway.

app/services/processors/blog_metrics_updater.rb
```diff
@@ -0,0 +1,20 @@
...
+  class BlogMetricsUpdater < BaseService
+    def call
+      update_blog_post_visits_metrics
+      update_technologies_visits_metrics
+    end
+
+    private
+
+    def update_blog_post_visits_metrics
+      BlogPost.find_each(batch_size: 50) do |blog_post|
```
hdamico, line 11: You could add lazy here to improve the perfomance of the query, it adds a limit to the query.

app/services/processors/blog_technology_views_updater.rb
```diff
@@ -0,0 +1,41 @@
...
+        update_technology_visits_metric(technology_id, visits, timestamp)
+      end
+    end
+
+    private
+
+    def blog_views_by_technology_and_month
+      Metric.where(name: Metric.names[:blog_visits], ownable_type: BlogPost.to_s)
+            .where('value_timestamp >= ?', latest_technology_metrics_updated_at)
+            .joins('JOIN blog_posts on metrics.ownable_id = blog_posts.id')
```
hdamico, line 15: This was made like this because of the polymorphic association right?

SandroDamilano: Yes! I had to write it that way because of that, I couldn't find any other way to do it.

app/services/processors/blog_metrics_updater.rb
SandroDamilano: Where exactly?
`BlogPost.lazy.find_each(batch_size: 50)` throws a `NoMethodError`
`BlogPost.find_each(batch_size: 50).lazy` seems to have no impact on the DB, thus making the tests fail
I didn't know that method, so I'm not sure how to use it 😅

hdamico: It should be something like this: `BlogPost.find_each(batch_size: 50).lazy.each`

SandroDamilano: All right, I'll see why it's making the tests fail then. Thanks!

PR #177, opened 2020-05-22
Rescue when a team is requested as reviewer

hdamico:


hdamico:
> Just a comment, but if we don't care about that case maybe it shouldn't be an exception

I would like to keep record on this to fix it in the future.

PR #180, opened 2020-05-26
[HOTFIX] Active Admin js not loaded in Production

hvilloria:
## What does this PR do?

* resolves error in production which does not load active admin file.


hvilloria:
> Fixed here: #179
got it, gonna close it then!

PR #184, opened 2020-05-26
[EM-1] Project Details from sidebar

hvilloria:
## What does this PR do?

* implements [Select2](https://select2.org/)
* refactors the sidebar to use filters depending on the layout displayed

* WIP: the merge time is should be implemented on this filtering

Resolves: [Jira](https://rootstrap.atlassian.net/browse/EM-1)


app/controllers/projects_controller.rb
```diff
@@ -21,7 +21,7 @@ def review_turnaround
   private
 
   def project_id
-    params[:project_id]
+    Project.find_by(name: params[:metric][:project_name])
```
santiagovidal, line 24: We should memoize this because we're calling it several times.

db/seeds.rb
```diff
@@ -6,6 +6,13 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 if Rails.env.development?
+  AdminUser.delete_all
+  User.delete_all
+  UsersProject.delete_all
+  Metric.delete_all
+  Technology.delete_all
+  Project.delete_all
```
hdamico, line 14: Why are we doing this?

hvilloria: to keep the consistency in dev-database, doing a delete from all the models avoid duplication/validation errors from the models previously created, every time we run a seed it just generate the current data and doesn't get mixed with previous data

hdamico: Wouldn't be better to just make a `rake db:reset`?

hvilloria: you are right, better a `reset`. going to remove it 👍

app/services/queries/daily_metrics.rb
```diff
@@ -3,7 +3,7 @@ class DailyMetrics < BaseQueryMetric
     FROM = 14.days
     INTERVAL = 'daily'.freeze
 
-    def initialize(project_id, metric_name)
+    def initialize(project_id:, metric_name:)
```
horacio, line 6: The constructor signature for both the daily and weekly queries went back and forth between keyword args, to positional args, to keyword args again. 😆 Nevertheless, we do need keyword args in this case, right? 🙌

hvilloria: jajaj yes, this was because of the changes that came from develop, this ticket has work from hernan too, so we had to merge our changes 😄

horacio: Awesome, thanks!

PR #187, opened 2020-05-27
Add chart for technology visits

SandroDamilano:
## What does this PR do?
It creates a chart for blog visits by technology metrics. The chart looks like this:

image

It's worth noting that technologies stats can be viewed separately by turning off the ones that are not needed (and the scale of the chart resizes accordingly):

image

### DIFF
--- a/app/services/blog_metric_chart_builder.rb
+++ b/app/services/blog_metric_chart_builder.rb
@@ -1,0 +1,18 @@
+class BlogMetricChartBuilder
+  def technology_visits
+    Technology.all.map do |technology|
+      {
+        name: technology.name.titlecase,
+        data: metrics_data_for(technology)
+      }
+    end
+  end
+
+  private
+
+  def metrics_data_for(technology)
+    technology.metrics.order(:value_timestamp).each.inject({}) do |hash, metric|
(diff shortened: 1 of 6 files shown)


app/services/blog_metric_chart_builder.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  private
+
+  def metrics_data_for(technology)
+    technology.metrics.order(:value_timestamp).each.inject({}) do |hash, metric|
+      hash.merge!(metric.value_timestamp.strftime('%B') => metric.value)
+    end
+  end
+end
```
hvilloria, line 18: now there is a class for chartkick data, we should move later this code there, what do you think? maybe we should create a ticket to do it

SandroDamilano: I agree that we should look for common logic between the two.
But I should note that in my first attempt I tried to use that other one and there were too many setbacks. Basically the `name` of the entity and how both the `value` and the `value_timestamp` are used is fairly different, so the common logic may result just in building the skeleton:
```
[{
  name: block_to_get_entity_name,
  data: {
    block_to_parse_value_timestamp => block_to_parse_value,
    ...
  }
},
...
]
```
Which I don't know how it compares to just having different ones

PR #192, opened 2020-05-29
[EM-1.1] Basic wireframe complete

hvilloria:
## What does this PR do?

* completed v1 wireframe
* root routes now goes to the metrics page

Screen Shot 2020-05-29 at 1 28 31 PM

### Extra:
* changes the dot use by pound in specs, rspec standard.


app/views/projects/user_project_metric.erb
```diff
@@ -1,19 +1,21 @@
...
-<%end%>
-<div class='col-md-10 p-3'>
-  <%= line_chart @metrics, ytitle: "Time as hours", suffix: " hours" %>
+<div class="col-md-10">
+  <%= render 'shared/nav-filter' do |f|%>
+      <%= f.input :period,
+      wrapper_html: { class: 'period-selection' },
+      input_html: { class: 'margin-left' },
+      required: false,
+      collection: Metric.intervals.to_a,
```
santiagovidal, line 7: Can we just show daily and weekly? Because monthly and all times will give an error until we implement them.

app/views/shared/_sidebar.html.erb
```diff
@@ -1,17 +1,21 @@
...
-          <%= yield f if block_given? %>
-      <% end %>
+      <label for="project_name">Project Name:</label>
+      <select name="project_name" class="project-selection" form='nav-filter-form'>
+        <%if params[:project_name].blank? %>
+          <option selected value disabled>-select option-</option>
+        <%else%>
+          <option value disabled>-select option-</option>
+        <%end%>
+        <% Project.pluck(:name).each do |name|%>
```
santiagovidal, line 11: Perhaps this can be loaded somewhere in a controller.

app/controllers/projects_controller.rb
```diff
@@ -2,16 +2,23 @@ class ProjectsController < ApplicationController
...
-      project_id: project_id,
-      metric_name: metric_params[:name]
-    )
+    period = metric_params[:period]
+    @review_turnaround = Queries::BaseQueryMetric.determinate_metric_period(period)
+                                                 .call(
+                                                   project_id: project_id,
+                                                   metric_name: 'review_turnaround'
+                                                 )
+    @merge_time = Queries::BaseQueryMetric.determinate_metric_period(period)
```
santiagovidal, line 11: Are we calling `determinate_metric_period` everytime we use the `call` method? If so, perhaps that code can go inside `call`.

app/views/projects/user_project_metric.erb
hvilloria: doing it!

app/views/shared/_sidebar.html.erb
hvilloria: how about a view helper? what do you think?

app/controllers/projects_controller.rb
hvilloria: roger that

app/views/shared/_sidebar.html.erb
santiagovidal: Yeah, could be

app/helpers/projects_names_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ProjectsNamesHelper
+  def projects_names
```
santib, line 2:
```suggestion
  def all_projects_names
```

app/javascript/modules/graph/index.js
```diff
@@ -1,34 +1,34 @@
...
+  let navFilterForm = elementSelector('nav-filter');
 
-  sidebarForm.onchange = function () {
-    let projectSelect = selectInputObject('project');
-    let periodSelect = selectInputObject('period');
-    let metricNameSelect = selectInputObject('metric-name');
+  sidebarProjectSelectionInput.onchange = () => {
+    const periodSelected = document.getElementById('metric_period');
+    if (periodSelected.selectedIndex === 0) {
+      periodSelected.selectedIndex = 1
```
santib, line 17:
```suggestion
      periodSelected.selectedIndex = 1;
```

app/javascript/modules/graph/index.js
```diff
@@ -1,34 +1,34 @@
...
 
-    if (noOptionSelected(projectSelect.optionSelected)) {
-      const url = window.location.href.split('?')[0]
-      window.location.assign(url);
-    } else {
-      noOptionSelected(periodSelect.optionSelected) && (periodSelect.element.selectedIndex = 1)
-      noOptionSelected(metricNameSelect.optionSelected) && (metricNameSelect.element.selectedIndex = 1)
+export const handleChangeNavForm = () => {
+  const navFilterForm = elementSelector('nav-filter');
+  navFilterForm.onchange = function () {
```
santib, line 25:
```suggestion
  navFilterForm.onchange = () => {
```

app/javascript/modules/graph/index.js
```diff
@@ -1,34 +1,34 @@
...
+    let sidebarProjectSelectionInput = elementSelector('project-selection');
+    if (sidebarProjectSelectionInput.selectedIndex !== 0){
       this.submit();
+    } else {
+      const url = window.location.href.split('?')[0];
+      window.location.assign(url);
```
santib, line 31: for a future PR, but if you can recognize from the controller that there is project selected and redirect to the clean url, then that's better

app/views/shared/_sidebar.html.erb
```diff
@@ -1,17 +1,21 @@
...
-            selected: value_from_query_param(:project_name),
-            include_blank: 'no option selected'
-          %>
-          <%= yield f if block_given? %>
-      <% end %>
+      <label for="project_name">Project Name:</label>
+      <select name="project_name" class="project-selection" form='nav-filter-form'>
+        <%if params[:project_name].blank? %>
+          <option selected value disabled>-select option-</option>
+        <%else%>
```
santib, line 8: use spaces like `<% else %>`

spec/services/processors/blog_posts_full_updater_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 describe Processors::BlogPostsFullUpdater do
-  describe '.call' do
+  describe '#call' do
```
santib, line 4: why this change? you are calling `Processors::BlogPostsFullUpdater.call` so that's a class method an `.call` is correct

app/views/projects/user_project_metric.erb
```diff
@@ -1,19 +1,21 @@
...
+<div class="col-md-10">
+  <%= render 'shared/nav-filter' do |f|%>
+      <%= f.input :period,
+      wrapper_html: { class: 'period-selection' },
+      input_html: { class: 'margin-left' },
+      required: false,
+      collection: [['Daily', 'daily'], ['Weekly', 'weekly']],
+      selected: value_from_query_param(:period),
+      prompt: '- select option -'
+
```
santib, line 10:
```suggestion
```

app/controllers/projects_controller.rb
```diff
@@ -2,16 +2,20 @@ class ProjectsController < ApplicationController
...
   end
 
   private
 
   def project_id
-    @project_id ||= Project.find_by(name: metric_params[:project_name]).id
+    @project_id ||= Project.find_by(name: params[:project_name]).id
   end
 
   def metric_params
```
hdamico, line 21: Can we move the hole controller to UsersProjectController? And keep the URL like this `/users_project/metrics` ?
Also, I am working on Project metrics, so I will need to create another method called `project_metrics` which I found a little redundant.

app/javascript/modules/graph/index.js
hvilloria: @santib reason of the function expression is to have `this` pointing to the form instead of the window

app/controllers/projects_controller.rb
hvilloria: of course, few minutes and i will upload the changes

app/views/layouts/_navigation.html.erb
```diff
@@ -1,6 +1,6 @@
 <nav class="nav nav-style nav-txt-color p-2 nav-items-container">
   <span class="navbar-brand nav-txt-color">Engineering Metrics</span>
-  <a class="nav-link active nav-txt-color" href="/projects/user_project_metric">Development Metrics</a>
+  <a class="nav-link active nav-txt-color" href="/users_projects/metrics">Development Metrics</a>
```
santiagovidal, line 3: It would be better to use the route helper

app/controllers/users_projects_controller.rb
```diff
@@ -0,0 +1,24 @@
+class UsersProjectsController < ApplicationController
+  def metrics
```
santib, line 2: for the future, it'd be great if our controllers are RESTful. e.g. for this case I'd have
```ruby
module Metrics
  class UsersProjectsController < ApplicationController
    def index
```

http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

PR #195, opened 2020-06-01
New tests for review turnaround per user project

hdamico:
## What does this PR do?

Adds two new tests, first one is the case when there are more that one review from the same user in the same pull request it should get the first review based on the opened_at date, the second one is when there are more than one review in a pull request but by different user it should create two metrics.


spec/services/metrics/review_turnaround/per_user_project_spec.rb
```diff
@@ -21,7 +22,7 @@
         let!(:review) do
           create(:review,
                  pull_request: pull_request,
-                 opened_at: Time.zone.now + 30.minutes,
+                 opened_at: current_time + 30.minutes,
```
santib, line 25: `30.minutes.from_now(current_time)`

PR #196, opened 2020-06-01
[EM-43] Allow blog metrics to be fully updated

SandroDamilano:
## What does this PR do?
Not all blog metrics need to be updated all the time; in a regular use, we should only update those metrics corresponding to the current month (and maybe the last one as well in some cases). Visits for previous months won't ever change.

But still there are certain circumstances where we do need to update all of them (for example, when a blog post changes its technology).

For those occasions, our `BlogMetricsUpdater` should be able to be run for a _full_ update as well as a _partial_ one.

Story: https://rootstrap.atlassian.net/browse/EM-43

### DIFF
--- a/app/services/processors/blog_technology_views_full_updater.rb
+++ b/app/services/processors/blog_technology_views_full_updater.rb
@@ -3,7 +3,7 @@ module Processors
     private
 
     def metrics_by_timestamp
-      Metric
+      Metric.all
     end
   end
 end
--- a/app/services/processors/blog_technology_views_full_updater.rb
+++ b/app/services/processors/blog_technology_views_full_updater.rb
@@ -1,0 +1,9 @@
+module Processors
+  class BlogTechnologyViewsFullUpdater < BlogTechnologyViewsUpdater
+    private
+
+    def metrics_by_timestamp
+      Metric
+    end
+  end
+end
--- a/spec/services/processors/blog_technology_views_updater_spec.rb
+++ b/spec/services/processors/blog_technology_views_updater_spec.rb
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-describe Processors::BlogTechnologyViewsUpdater do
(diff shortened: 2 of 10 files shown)


app/services/processors/blog_technology_views_full_updater.rb
```diff
@@ -0,0 +1,9 @@
+module Processors
+  class BlogTechnologyViewsFullUpdater < BlogTechnologyViewsUpdater
+    private
+
+    def metrics_by_timestamp
+      Metric
```
santiagovidal, line 6: Should this return `Metric.all`, so that it is consistent with what the other `metrics_by_timestamp` method returns (an Arel)

spec/services/processors/blog_technology_views_updater_spec.rb
```diff
@@ -87,35 +89,7 @@
...
-            :metric,
-            name: Metric.names[:blog_visits],
-            interval: Metric.intervals[:monthly],
-            value: technology_visits,
-            value_timestamp: metric_timestamp,
-            ownable: technology
-          )
-        end
-
-        it 'just tries to update the last of them' do
```
santib, line 114: why is this test removed?

app/services/processors/blog_technology_views_full_updater.rb
SandroDamilano: Sure!

spec/services/processors/blog_technology_views_updater_spec.rb
SandroDamilano: It was moved actually, to `blog_technology_views_partial_updater_spec.rb` (https://github.com/rootstrap/rs-code-review-metrics/pull/196/files#diff-ce642b8ab10a78e0f9b5844865b268bbR48)
This behaviour is not related anymore to this class, it depends on the subclass being used (`BlogTechnologyViewsPartialUpdater ` or `BlogTechnologyViewsFullUpdater `)

PR #197, opened 2020-06-02
Schedule Blog jobs

SandroDamilano:
## What does this PR do?
It schedules three main jobs to update the blog metrics:
- `BlogPostsPartialUpdaterJob`: Runs once a week, on Wednesdays early morning. It adds newly published posts to our DB.
- `BlogPostsFullUpdaterJob`: Runs once per month (on the 1st). It updates all blog posts from scratch (looking for updated titles, links, tags or whatever new info that we should be aware of).
- `BlogMetricsPartialUpdaterJob`: Runs every day, early morning. It updates the metrics with the latest info.
- `BlogMetricsFullUpdaterJob`: Runs right after `BlogPostsFullUpdaterJob`. It reflects those possible changes in the blog posts into the metrics.

### DIFF
--- a/config/schedule.yml
+++ b/config/schedule.yml
@@ -9,3 +9,19 @@ weekly_metrics:
 merge_time_generator:
   cron: '00 23 * * *'
   class: 'ProcessMergeTimeMetricsJob'
+
+blog_posts_partial_update:
+  cron: '00 4 * * 3'
+  class: 'BlogPostsPartialUpdaterJob'
+
+blog_posts_full_update:
+  cron: '00 2 1 * *'
+  class: 'BlogPostsFullUpdaterJob'
+
+blog_metrics_partial_update:
+  cron: '30 4 * * *'
+  class: 'BlogMetricsPartialUpdaterJob'
+
+blog_metrics_full_update:
+  cron: '00 3 1 * *'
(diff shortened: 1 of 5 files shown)


config/schedule.yml
```diff
@@ -9,3 +9,19 @@ weekly_metrics:
...
+  cron: '00 2 1 * *'
+  class: 'BlogPostsFullUpdaterJob'
+
+blog_metrics_partial_update:
+  cron: '30 4 * * *'
+  class: 'BlogMetricsPartialUpdaterJob'
+
+blog_metrics_full_update:
+  cron: '00 3 1 * *'
+  class: 'BlogMetricsFullUpdaterJob'
```
hdamico, line 27: Add active_job: true at the end of every cron job, eg: 
```
blog_posts_partial_update:
  cron: '00 4 * * 3'
  class: 'BlogPostsPartialUpdaterJob'
  active_job: true
```
And as a boy scout add that to the other ones :)

SandroDamilano: Sure, no problem!
Why is that for?

hdamico: Sidekiq-cron gem documentation

image

We are using active job interface in our proyect.

PR #210, opened 2020-06-04
[EM-23.1] Create blog post count metrics

SandroDamilano:
## What does this PR do?
It adds the `blog_post_count` metrics
- `groupdate` gem is added
- The `blog_post_count` value is added to the `metric_name` type in DB
- A `BlogPostCountMetricsUpdater` processor is added (along with its _full_ and _partial_ variations). It's in charge of generating the corresponding metrics.
- These processors executions are added to the `BlogMetricsUpdater` process

Bonus: I added a test for `blog_technology_views_partial_updater_spec.rb` (the case was contemplated but not tested)


app/services/processors/blog_post_count_metrics_updater.rb
```diff
@@ -0,0 +1,53 @@
+module Processors
+  class BlogPostCountMetricsUpdater < BaseService
+    def call
+      Technology.all.each do |technology|
+        new_blog_posts_by_month_for(technology).each do |month, new_blog_post_count|
```
santiagovidal, line 5: `find_each` would be preferable maybe

SandroDamilano: Where would you add it?
`new_blog_posts_by_month_for(technology)` returns a hash, not a relation, so we can't use it there.
And there should be very few records in `Technology` (not more than 10), so I don't think iterating though them in batches would do much.

hdamico: It would be something like this: `Technology.find_each do |technology|`, gets records in batches and not all at once. By default is a batch of 1000 records

SandroDamilano: But if `Technology` won't have more than 10 records, 15 at most, would it make any difference?
(I'm not opposed to the change, I'm just trying to understand the benefit)

santiagovidal: I thought it was an Arel, in that case ignore my comment

db/migrate/20200602181502_add_blog_post_count_to_metric_name_type.rb
```diff
@@ -0,0 +1,16 @@
+class AddBlogPostCountToMetricNameType < ActiveRecord::Migration[6.0]
+  def up
```
hdamico, line 2: I believe you should add `disable_ddl_transaction!` 
WDYT @mrubi-rootstrap ?

mrubi-rootstrap: If the migration fails with enums in Postgres that fixes it, yes

SandroDamilano: Done! Thank you!

PR #211, opened 2020-06-05
[EM-23.2] Show blog post count metrics

SandroDamilano:
## What does this PR do?
It adds the chart showing blog posts count per technology in the Blog Metrics tab

image


hdamico:
Just a suggestion:
  Wouldn't be better having a folders builders/blog_metric_chart with technology_visits.rb and technology_blog_post_count.rb, which both inherit from a class called Base or something like inside of builders/blog_metric_chart too, which inherit from base_service, so we can keep the convention of calling services with `call` method. And in the controller you can do something like this: `Builders::BlogMetricChart::NewTechnologyVisits.call` and `Builders::BlogMetricChart::TechnologyBlogPostCount.call`

SandroDamilano:
@hdamico Oh, yeah, I see! I'll restructure it that way, then

PR #214, opened 2020-06-08
Feature/em 15 integrate code climate

mrubi-rootstrap:
## What does this PR do?

It integrates [CodeClimate into the project view](https://rootstrap.atlassian.net/jira/software/projects/EM/boards/42?selectedIssue=EM-15) 

It does not implement the population of the CodeClimate data which is to be implemented in a different task

### DIFF
--- a/app/services/metrics/code_climate_summary_retriever.rb
+++ b/app/services/metrics/code_climate_summary_retriever.rb
@@ -7,11 +7,7 @@ module Queries
     end
 
     def call
-      {
-        code_climate_rate: 'A',
-        code_climate_invalid_issues_count: 1,
-        code_climate_wont_fix_issues_count: 2
-      }
+      CodeClimateProjectMetric.joins(:project).where(projects: { id: project_id }).first
     end
   end
 end
--- a/app/services/metrics/code_climate_summary_retriever.rb
+++ b/app/services/metrics/code_climate_summary_retriever.rb
@@ -1,0 +1,17 @@
+module Queries
+  class CodeClimateSummaryRetriever < BaseService
+    attr_reader :project_id
+
+    def initialize(project_id)
+      @project_id = project_id
+    end
+
+    def call
+      {
+        code_climate_rate: 'A',
+        code_climate_invalid_issues_count: 1,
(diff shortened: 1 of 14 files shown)


db/structure.sql
```diff
@@ -106,7 +106,7 @@ CREATE TYPE public.review_state AS ENUM (
 
 SET default_tablespace = '';
 
-SET default_table_access_method = heap;
+SET default_with_oids = false;
```
hdamico, line 109: What does this do?

app/services/metrics/code_climate_summary_retriever.rb
```diff
@@ -0,0 +1,13 @@
+module Metrics
+  class CodeClimateSummaryRetriever < BaseService
+    attr_reader :project_id
+
+    def initialize(project_id)
+      @project_id = project_id
+    end
+
+    def call
+      CodeClimateProjectMetric.joins(:project).where(projects: { id: project_id }).first
```
hdamico, line 10: Could we use `find_by` right?
`CodeClimateProjectMetric.joins(:project).find_by(projects: { id: project_id })`

db/structure.sql
mrubi-rootstrap: I couldn't tell. This file is generated by Rails every time I recreate the migrations.

hdamico: I am on your branch and I ran the migrations on my local and I don't have that 🤔 
image

app/services/metrics/code_climate_summary_retriever.rb
mrubi-rootstrap: Indeed.
Fixed it. Thanks!

db/structure.sql
horacio: This is most likely a DB version thing.

@mrubi-rootstrap which version of Postgres do you have?

mrubi-rootstrap: It seems like a Postgres setting, yes.
I am using postgres:11 in a docker container, will look for setting that value to match the previous value

mrubi-rootstrap: In the mean time I reverted it to its previous value

PR #215, opened 2020-06-09
[EM-50] - Feature/per project view

hdamico:
## What does this PR do?

Jun-09-2020 14-09-00

Jira: [EM-50](https://rootstrap.atlassian.net/browse/EM-50)


app/javascript/packs/togglePerUserProject.js
```diff
@@ -0,0 +1,14 @@
+$(document).ready(function() {
```
santiagovidal, line 1: I think we have to use the turbolinks listener here. @santib WDYT?

app/views/metrics/index.erb
```diff
@@ -0,0 +1,16 @@
...
+      wrapper_html: { class: 'period-selection' },
+      input_html: { class: 'margin-left' },
+      required: false,
+      collection: [['Daily', 'daily'], ['Weekly', 'weekly']],
+      selected: value_from_query_param(:period),
+      prompt: '- select option -'
+    %>
+  <% end %>
+  <%= render template: "projects/review_turnaround", locals: { review_turnaround: @review_turnaround } %>
+  <button type="button" class="btn btn-secondary details-button">View details</button>
```
santiagovidal, line 13: I have an alternative solution here:

We write a generic partial that contains:
- The project metric
- The button to see details
- The user-project metric

And we write it in a generic way, so that merge time (and other future metric) use it as well, passing the correct arguments each time it is rendered.

I'm open to opinions here

hdamico: I like it! I will put the metrics and the button on a review turnaround partial

PR #216, opened 2020-06-09
Show growth mom metrics

SandroDamilano:
## What does this PR do?
It adds the visits growth month over month per technology chart in the Blog Metrics tab.
These metrics are being calculated on demand, instead of being pre-calculated.

image


app/services/technology_growth_mom_calculator.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  attr_reader :metric_name, :months
+
+  def growth_mom_for(technology)
+    metrics = metrics_for(technology)
+
+    metrics.each.with_index.inject({}) do |hash, metric_with_index|
+      metric, index = metric_with_index
+      previous_metric = metrics[index - 1]
+      next(hash) if index.zero? || previous_metric.value.zero?
```
horacio, line 26: Would

```suggestion
      next if index.zero? || previous_metric.value.zero?
```

work too?

SandroDamilano: It wouldn't, I've tried it. It seems that it needs to keep passing the seed on to the next iteration.

PR #220, opened 2020-06-11
[Story] Code Owners list

hvilloria:
## What does this PR do?

* shows the code owner list.

Screen Shot 2020-06-11 at 5 03 51 PM

Issue: [Jira](https://rootstrap.atlassian.net/browse/EM-2)


app/services/code_owners_service.rb
```diff
@@ -0,0 +1,34 @@
+class CodeOwnersService < BaseService
+  def call
+    Project.all.find_each do |project|
```
horacio, line 3:
```suggestion
    Project.find_each do |project|
```

app/services/code_owners_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      content_file = GithubReposApi.new(project.name).get_content_from_file('CODEOWNERS')
+      next if content_file.empty?
+
+      code_owners_names = build_code_owners_array(content_file)
+      remove_old_code_owners(code_owners_names, project)
+      add_new_code_owners(code_owners_names, project)
```
horacio, line 9: Maybe we can benefit from using instance variables instead of passing around parameters, WDYT?

spec/models/code_owner_project_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+#  fk_rails_...  (project_id => projects.id)
+#  fk_rails_...  (user_id => users.id)
+#
+
+require 'rails_helper'
+
+RSpec.describe CodeOwnerProject, type: :model do
+  it { is_expected.to belong_to(:project) }
+  it { is_expected.to belong_to(:project) }
```
horacio, line 26: Duped line!

app/services/code_owners_service.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def build_code_owners_array(content_file)
+    content_file
+      .split('*')
+      .second
+      .remove(' ')
+      .strip
+      .split('@')
+      .reject(&:blank?)
```
horacio, line 32: WDYT about something like...

```ruby
content_file.lines.flat_map do |line|
  line.scan(%r{@(\w+)}).flatten
end
```

hvilloria: 👏 👏  awesome! changing it

app/services/code_owners_service.rb
santib: could be a good idea, but that would mean having a new class. It wouldn't be a good idea to be setting instance variables inside a loop

spec/services/code_owners_service_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+          it 'adds code_owners that are not in the file' do
+            allow_any_instance_of(GithubReposApi)
+              .to receive(:get_content_from_file) { updated_code_owners_file }
+            described_class.call
+            expect(project.code_owners.include?(hosward_code_owner)).to be(false)
+          end
+
+          it 'adds code_owners that are in the file' do
+            allow_any_instance_of(GithubReposApi)
```
santib, line 52: why instead of mocking the class don't you use `WebMock` such as you did here spec/support/helpers/github_api_mocker.rb ? I think that'd be better

app/services/code_owners_service.rb
horacio: you're absolutely right! man I got so confused.

app/external_services/github_repos_api.rb
```diff
@@ -0,0 +1,15 @@
...
+  def initialize(project_name)
+    @project_name = project_name
+  end
+
+  def get_content_from_file(file_name)
+    response = Faraday.get("#{URL}/#{@project_name}/contents/#{file_name}", {},
+                           'Accept' => 'application/vnd.github.v3.raw')
+    return response.body if response.success?
+
+    {}
```
mrubi-rootstrap, line 13: In this case would invert the early return to

```ruby
return {} unless response.success?
response.body
```

returning early for anything but the happy path mentioned in the method name.

Or since this conditional is a straight forward accessor with not logic maybe

```ruby
response.success? ? response.body : {}
```

app/services/code_owners_service.rb
```diff
@@ -0,0 +1,30 @@
+class CodeOwnersService < BaseService
+  def call
+    Project.find_each do |project|
+      content_file = GithubReposApi.new(project.name).get_content_from_file('CODEOWNERS')
+      next if content_file.empty?
+
+      code_owners_names = build_code_owners_array(content_file)
```
mrubi-rootstrap, line 7: A think a `class CodeOwnersFile` would simplify the code a bit

```ruby
code_owners_file = GithubReposApi.new(project.name).code_owners_file
code_owners_names = code_owners_file.code_owner_names
```

app/services/code_owners_service.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def remove_old_code_owners(code_owners_names, project)
+    project.code_owners.delete(User.where.not(login: code_owners_names))
+  end
+
+  def build_code_owners_array(content_file)
+    content_file.lines.flat_map do |line|
+      line.scan(%r{@(\w+)}).flatten
+    end
```
mrubi-rootstrap, line 28: I would extract this code to a class `CodeOwnersFile` with a method `def code_owner_names`

app/views/development_metrics/index.erb
```diff
@@ -10,7 +10,20 @@
...
+        <ul>
+          <% @code_owners.each do |code_owner|%>
+            <li>
+              <span class="badge badge-secondary"><%= code_owner %></span>
+            </li>
+          <%end%>
+        </ul>
+      <%end%>
+    </div>
+  </div>
```
mrubi-rootstrap, line 26: What do you think of extracting this div element to a partial like

```ruby
 <%= render 'code_owners_turnaround', review_turnaround: @code_owners %>
```

?

spec/services/code_owners_service_spec.rb
mrubi-rootstrap: I second that. I started using it and so far I think it's a bit more clear than vcr

spec/support/helpers/github_api_mocker.rb
```diff
@@ -0,0 +1,13 @@
...
+      .to_return(body: not_found_body_from_github, status: 404)
+  end
+
+  def not_found_body_from_github
+    {
+      'message': 'Not Found',
+      'documentation_url': 'https://developer.github.com/v3/repos/contents/#get-contents'
+    }.to_json
+  end
+end
```
mrubi-rootstrap, line 13: I think this is awesome!

app/models/code_owner_project.rb
```diff
@@ -0,0 +1,25 @@
...
+#  index_code_owner_projects_on_project_id  (project_id)
+#  index_code_owner_projects_on_user_id     (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (project_id => projects.id)
+#  fk_rails_...  (user_id => users.id)
+#
+
+class CodeOwnerProject < ApplicationRecord
```
SandroDamilano, line 22: Absolutely not against this solution, but have you considered using the `UsersProject` entity for this? Maybe just adding a boolean column `code_owner` could have worked.
Again, just an observation that may have come too late, not that we should go for it now.

app/external_services/github_repos_api.rb
```diff
@@ -0,0 +1,15 @@
+class GithubReposApi
```
santiagovidal, line 1: Any reason why we can't have this service under `services`, with all the other service classes?

app/services/code_owners_service.rb
```diff
@@ -0,0 +1,30 @@
...
+      code_owners_names = build_code_owners_array(content_file)
+      remove_old_code_owners(code_owners_names, project)
+      add_new_code_owners(code_owners_names, project)
+    end
+  end
+
+  def add_new_code_owners(code_owners_names, project)
+    current_code_owners = project.code_owners
+    code_owners_names.each do |code_owner_name|
+      user = User.find_by(login: code_owner_name)
```
santiagovidal, line 16: What if the user hasn't been created yet? Is it possible? Perhaps `find_or_create_by` will be better

app/services/code_owners_service.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  def remove_old_code_owners(code_owners_names, project)
+    project.code_owners.delete(User.where.not(login: code_owners_names))
+  end
+
+  def build_code_owners_array(content_file)
+    content_file.lines.flat_map do |line|
+      line.scan(%r{@(\w+)}).flatten
```
santiagovidal, line 27: Some codeowners files have examples in them, are you taking that into account?

Take a look at https://github.com/rootstrap/herddle-api/blob/develop/CODEOWNERS

spec/external_services/github_repos_api_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe GithubReposApi do
+  describe '#get_content_of_file' do
+    context 'when the project or the file is not found' do
```
santiagovidal, line 5: It would be good to have a context where the file **is** found.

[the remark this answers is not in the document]
hvilloria: yes, it is already tested in the code owners service 👍

app/services/code_owners_service.rb
```diff
@@ -0,0 +1,30 @@
...
+      remove_old_code_owners(code_owners_names, project)
+      add_new_code_owners(code_owners_names, project)
+    end
+  end
+
+  def add_new_code_owners(code_owners_names, project)
+    current_code_owners = project.code_owners
+    code_owners_names.each do |code_owner_name|
+      user = User.find_by(login: code_owner_name)
+      current_code_owners << user unless current_code_owners.include?(user)
```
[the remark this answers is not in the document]
hvilloria: yes, there it is creating the association record with ids from the project and user

app/external_services/github_repos_api.rb
hvilloria: the main reason is that all the services seems to have a pattern, they extends from `BaseService` and its entry points are the method `call`, for classes like this is probably that we will need to use methods for differents type of requests (GET some uri, POST another, etc).

app/services/code_owners_service.rb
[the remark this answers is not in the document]
hvilloria: i will test it with a real content file to cover those cases 👍

app/services/code_owners_service.rb
hvilloria: in that flow we do not have all the attributes that are needed to create a user, in theory the cases where the user doesn't exist yet should be few..  we could do a `next` to skip it for the moment until a PR, Review or Review Request is created. what do you think?

app/services/code_owners/file_handler.rb
```diff
@@ -0,0 +1,20 @@
...
+    # \*          - Literal
+    # ?<user>     - create a capturing group named "user"
+    # \s+@\w+     - main regexp filter: all @users separated by spaces
+    # (\g<user>)? - where "user" is the name of a capturing group
+
+    def call
+      @content_file.scan(/\*(?<user>\s+@\w+(\g<user>)?)/).flat_map do |one_line_names_string|
+        one_line_names_string.first.strip.split(/\s/).map do |name|
+          name.scan(/@(\w+)/).flatten.first
+        end
```
hvilloria, line 16: guys, need some help here 🙏  i don't know if there is another way. @horacio @santib @santiagovidal

santiagovidal: Looks good to me

app/services/code_owners/file_handler.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    # \*          - Literal
+    # ?<user>     - create a capturing group named "user"
+    # \s+@\w+     - main regexp filter: all @users separated by spaces
+    # (\g<user>)? - where "user" is the name of a capturing group
+
+    def call
+      @content_file.scan(/\*(?<user>\s+@\w+(\g<user>)?)/).flat_map do |one_line_names_string|
+        one_line_names_string.first.strip.split(/\s/).map do |name|
+          name.scan(/@(\w+)/).flatten.first
```
horacio, line 15: Looks good, Hoswi! I think you could do

```suggestion
          name.sub!('@', '')
```

But I think it's understandable as it is.

PR #222, opened 2020-06-16
[EM-52] Add totals to blog metrics

SandroDamilano:
## What does this PR do?
It calculates the totals for each blog metric and adds them to the charts

image

## Link to Story
https://rootstrap.atlassian.net/jira/software/projects/EM/boards/42?selectedIssue=EM-52


app/services/builders/blog_metric_chart/growth_month_over_month.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      def process_data(metrics_data)
+        metrics_data.each.inject({}) do |hash, key_value_pair|
+          month, metric_value = key_value_pair
+          previous_metric_value = metrics_data[month.last_month]
+          next(hash) if previous_metric_value.nil? || previous_metric_value.zero?
+
+          hash.merge!(
+            month => growth_for(metric_value, previous_metric_value)
+          )
```
horacio, line 18: Maybe this could be written as... 

```suggestion
          hash[month] = growth_for(metric_value, previous_metric_value)
```

WDYT?

app/services/builders/blog_metric_chart/growth_month_over_month.rb
```diff
@@ -0,0 +1,27 @@
...
+    class GrowthMonthOverMonth < Builders::BlogMetricChart::Base
+      def initialize(months = 13)
+        super(months + 1)
+      end
+
+      private
+
+      def process_data(metrics_data)
+        metrics_data.each.inject({}) do |hash, key_value_pair|
+          month, metric_value = key_value_pair
```
horacio, line 12: Another way to accomplish the same could be... 

```suggestion
        metrics_data.each_with_object({}) do |hash, (month, metric_value)|
```

WDYT?

app/services/builders/blog_metric_chart/growth_month_over_month.rb
```diff
@@ -0,0 +1,27 @@
...
+          previous_metric_value = metrics_data[month.last_month]
+          next(hash) if previous_metric_value.nil? || previous_metric_value.zero?
```
horacio, line 14:
```suggestion
          next(hash) if previous_metric_value.to_i.zero?
```

app/services/builders/blog_metric_chart/growth_month_over_month.rb
SandroDamilano: Oh, right, I completely forgot I could write it that way. Love it!

app/services/builders/blog_metric_chart/growth_month_over_month.rb
SandroDamilano: I think we may loose some expressiveness with this change. But I'm not against it, I can do it

app/services/builders/blog_metric_chart/growth_month_over_month.rb
SandroDamilano: I thought I couldn't do that. Thank you!

PR #224, opened 2020-06-18
Per department metrics

hdamico:
## What does this PR do?

Jun-18-2020 12-26-28

Resolves [EM-18](https://rootstrap.atlassian.net/browse/EM-18) and [EM-17](https://rootstrap.atlassian.net/browse/EM-17)


app/models/project.rb
```diff
@@ -2,20 +2,31 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (department_id => departments.id)
 #
 
 class Project < ApplicationRecord
   enum lang: { ruby: 'ruby', python: 'python', nodejs: 'nodejs',
-               react: 'react', ios: 'ios', android: 'android',
+               ios: 'ios', android: 'android', reactnative: 'react_native',
```
hvilloria, line 25:
```suggestion
               ios: 'ios', android: 'android', react_native: 'react_native',
```

im not sure if it is a mistake or you did it for a reason

spec/controllers/development_metrics_controller_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe DevelopmentMetricsController, type: :controller do
```
hvilloria, line 3: maybe later we could start to use request instead, im not sure but i think @santib sayd it is gonna be deprecated

app/helpers/departments_names_helper.rb
```diff
@@ -0,0 +1,5 @@
+module DepartmentsNamesHelper
+  def all_departments_names
+    Department.pluck(:name).sort
+  end
+end
```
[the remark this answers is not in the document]
hdamico: Yes! I thought of doing something similar but I didn't want to add more stuffs to the PR. So go ahead with that!

app/models/project.rb
hdamico: I preffer to keep it with underscore, so in the future we can just do a humanize or something like that, and obtain ReactNative. Maybe we should do the same with nodejs and vuejs

app/helpers/departments_names_helper.rb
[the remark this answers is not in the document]
hvilloria: 👏

app/services/metrics/merge_time/per_department.rb
```diff
@@ -0,0 +1,42 @@
...
+        department_metrics_values.each do |department_name, metrics_values|
+          department_id = Department.find_by!(name: department_name)
+          turnaround = calculate_turnaround(metrics_values)
+          create_or_update_metric(department_id, Department.to_s, metric_interval,
+                                  turnaround, :review_turnaround)
+        end
+      end
+
+      def department_metrics_values
+        Department.names.keys.each_with_object({}) do |department_name, hash|
```
santiagovidal, line 24: Why not just iterate through the department objects, so that you don't have to find them in line 16

app/services/metrics/merge_time/per_department.rb
```diff
@@ -0,0 +1,42 @@
...
+          hash[department_name] = Metric.joins(project: :department)
+                                        .where(departments: { name: department_name })
+                                        .where(name: :merge_time)
+                                        .pluck(:value)
+          hash
+        end
+      end
+
+      def calculate_turnaround(metrics_values)
```
santiagovidal, line 33: This would be `calculate_average` right?

app/services/metrics/review_turnaround/per_department.rb
```diff
@@ -0,0 +1,44 @@
...
+              .pluck('departments.id', :value)
+              .each_with_object({}) do |(id, value), hash|
+                (hash[id] ||= []) << value
+                hash
+              end
+      end
+
+      def calculate_turnaround(metrics_values)
+        metrics_values_size = metrics_values.size
+        return metrics_values.first if metrics_values_size == 1
```
santiagovidal, line 34: We can remove this line and it won't change anything right? If `metrics_values_size` is 1, then `metrics_values.sum / metrics_values_size` will be `metrics_values.first`

app/views/development_metrics/merge_time/_project.erb
```diff
@@ -0,0 +1,5 @@
+<%= render partial: 'shared/line-chart',
```
santiagovidal, line 1: Should we rename this partial to something more generic? It is used for all types of metrics, not just `project` metrics right?

app/models/project.rb
```diff
@@ -36,4 +47,21 @@ class Project < ApplicationRecord
...
+  DEPARTMENT_LANGS = {
+    backend: %w[ruby python nodejs].freeze,
+    frontend: %w[react vuejs].freeze,
+    mobile: %w[ios android react_native].freeze
+  }.freeze
+
+  def associate_department
+    return unless department.nil? || lang_changed?
+
+    department_name = DEPARTMENT_LANGS.select { |_key, values| values.include?(lang) }.keys.first
```
mrubi-rootstrap, line 64: Maybe this line could be extracted to a method

```ruby
def find_lang_department
  DEPARTMENT_LANGS.select { |_key, values| values.include?(lang) }.keys.first
end
```

app/services/builders/chartkick/department_data.rb
```diff
@@ -0,0 +1,12 @@
+module Builders
+  module Chartkick
+    class DepartmentData < Builders::Chartkick::Base
```
mrubi-rootstrap, line 3: I would rename the class to `DepartmentDataRetriever` or to `DepartmentDataService` to make it clear that the object is not the DepartmentData itself but it returns DepartmentData

app/services/metrics/review_turnaround/per_department.rb
mrubi-rootstrap: I am not sure about `/` meaning between 2 integers. If it happens to be an integer division the remainder would be lost.
I agree with you but would use

```
metrics_values.sum / metrics_values_size.to_f
```

to ensure it's not an integer division

app/services/builders/chartkick/department_data.rb
hdamico: In fact is a data builder, I believe that `Builders::Chartkick::DepartmentData` is clear enough. WDYT @hvilloria ?

mrubi-rootstrap: Maybe you are right. I read only the class name but in its uses it will have the fully qualified name

app/services/metrics/merge_time/per_department.rb
hdamico: Fixed!

app/services/metrics/review_turnaround/per_department.rb
hdamico: @mrubi-rootstrap in this method we always receive the metric value, which is a decimal value, so never will be a integer value.

app/views/development_metrics/merge_time/_project.erb
hdamico: I'm not sure what you mean

app/services/metrics/review_turnaround/per_department.rb
mrubi-rootstrap: All right. Since that is something that depends on a parameter received I would still ensure the `metrics_values_size.to_f` to be a non integer Numeric or I would add a test to flag an error if the parameter received happens to be of a different type though. Just to make sure

app/services/metrics/base.rb
```diff
@@ -0,0 +1,28 @@
+module Metrics
+  class Base < BaseService
+    def calculate_avg(entities, metric_type)
```
santiagovidal, line 3: Perhaps we can rename this to `calculate_metrics_avg` or something like that, so it's not confused with the other method, `calculate_average`

app/views/development_metrics/merge_time/_project.erb
santiagovidal: Sorry, I realize now that it was misleading 😄 

I meant the name of the partial you are editing (`merge_time/_project.rb`). Should it still have that name, given that you are using it for department metrics?

app/controllers/development_metrics_controller.rb
```diff
@@ -1,25 +1,36 @@
...
-    @merge_time = metrics[:merge_time]
-    @code_climate = CodeClimateSummaryRetriever.call(project.id)
+    build_metrics(project.id, Project.to_s)
     @code_owners = project.code_owners.pluck(:login)
   end
 
+  def departments
+    return if metric_params.blank?
+
+    build_metrics(department_id, Department.to_s)
```
horacio, line 14: Another option: `Department.name`.

app/services/builders/chartkick/development_metrics.rb
```diff
@@ -12,19 +12,36 @@ def call
...
+          METRIC_NAMES.each_with_object({}) do |metric_name, hash|
+            hash[metric_name] = {
+              per_users_project: @period_metric_query.call(
+                entity_name: 'users_project', entity_id: @entity_id, metric_name: metric_name
+              ),
+              per_project: @period_metric_query.call(
+                entity_name: 'project', entity_id: @entity_id, metric_name: metric_name
+              )
+            }
+            hash
```
horacio, line 28: This isn't needed.

app/services/builders/chartkick/development_metrics.rb
```diff
@@ -12,19 +12,36 @@ def call
...
+        private
+
+        def build
+          METRIC_NAMES.each_with_object({}) do |metric_name, hash|
+            hash[metric_name] = {
+              per_department: @period_metric_query.call(
+                entity_name: 'department', entity_id: @entity_id, metric_name: metric_name
+              )
+            }
+            hash
```
horacio, line 43: This isn't needed

PR #225, opened 2020-06-18
[Story][Project list for codeowner]

hvilloria:
## What does this PR do?

* display a dropdown on sidebar to search users and shows the list of projects where they are code owners

[Jira](https://rootstrap.atlassian.net/jira/software/projects/EM/boards/42?selectedIssue=EM-59)

projects_code_owner


app/javascript/modules/sidebar/index.js
```diff
@@ -22,19 +22,28 @@ export const handleChangeSidebar = () => {
...
+  if (navFilterForm != null) {
+    navFilterForm.onchange = function () {
+      let sidebarProjectSelectionInput = elementSelector('project-selection');
+      if (sidebarProjectSelectionInput.selectedIndex !== 0){
+        this.submit();
+      } else {
+        const url = window.location.href.split('?')[0];
+        window.location.assign(url);
+      }
     }
```
hvilloria, line 34: @santiagovidal @horacio @santib i have one question here: when a navigate the navbar from to select the period is  not present in the next page view so this cause an error because it is trying to subscribe a `onchange` in a null object, is there a better way maybe to bind some js to specifics views or something like that? in order to avoid the `if navFilterForm != null`

santiagovidal: I'm not a JS guru, but I'm fine with your approach

spec/controllers/code_owners/projects_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe CodeOwners::ProjectsController, type: :request do
```
santiagovidal, line 3: If this is a `request` spec, why name it projects_controller_spec.rb 🤔

config/routes.rb
```diff
@@ -12,4 +12,7 @@
   post '/github_event_handler', to: 'webhook#handle'
   get '/development_metrics', to: 'development_metrics#index'
   get 'tech_blog', to: 'tech_blog#index'
+  namespace :code_owners do
```
santiagovidal, line 15: Can we name this route something basic like `/users/:id`

It's more RESTful and also, we will use this page for multiple user metrics, so it will be kind of the user's page.

spec/controllers/code_owners/projects_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      it 'returns no projects' do
+        expect(assigns(:projects)).to be_empty
+      end
+    end
+
+    context 'when projects has projects as code owner' do
+      before do
+        create_list(:project, projects_count)
+        Project.all.each { |project| project.code_owners << user }
+        get '/code_owners/projects', params: { code_owner: { user_id: user.id } }
```
santiagovidal, line 22: You can move this to a variable so as to not repeat it.

spec/controllers/code_owners/projects_controller_spec.rb
hvilloria: because as i think i understood, `type: :controller` is gonna be deprecated and should be changed by `type: : request`

app/javascript/modules/sidebar/index.js
horacio: Ditto. LGTM!

app/views/code_owners/projects/index.erb
```diff
@@ -0,0 +1,25 @@
...
+              class: 'link-none-style-black'
+          %>
+        </li>
+      <%end%>
+    </ul>
+  <%else%>
+    <p>
+      <br>
+      <strong>
+        <em>This user is not participating at any project as code owner</em>
```
horacio, line 21:
```suggestion
        <em>This user is not participating in any project as code owner</em>.
```

spec/controllers/code_owners/projects_controller_spec.rb
horacio: Actually, I think we don't need to pass this metadatum to `describe`, [as we already tell RSpec to infer its type](https://github.com/rootstrap/rs-code-review-metrics/blob/develop/spec/rails_helper.rb#L65).

You could move this file to `spec/requests` and rename it.

app/views/shared/user/_sidebar.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="col-md-2 container-height shadow-box sticky-top">
+  <div class="container-md">
+    <div class="row p-3">
+      <div class="btn-group btn-entity-group">
+        <%= button_to 'Project', projects_development_metrics_path, class: 'btn btn-secondary entity-btn', method: :get %>
```
santiagovidal, line 5: I see that this sidebar with these three buttons is repeated 3 times. Perhaps we reuse the same partial, and use a conditional to decide which select_tag to use. Not a blocker for this PR though

spec/requests/code_owner_projects.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe 'Code Owner Projects' do
```
SandroDamilano, line 3: Shouldn't this file name finish with `_spec`?

spec/requests/code_owner_projects.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      it_behaves_like 'controller index response'
+
+      it 'returns no projects' do
+        expect(assigns(:projects)).to be_empty
+      end
+    end
+
+    context 'when projects has projects as code owner' do
```
SandroDamilano, line 19: "projects has projects"? 😂

spec/requests/code_owner_projects.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe 'Code Owner Projects' do
+  describe '#index' do
+    let(:user) { create(:user) }
+    subject { get "/users/#{user.id}/projects" }
+    context 'when a user has not projects as code owner' do
+      before do
+        subject
+      end
```
SandroDamilano, line 10: I'm not too fond on doing the _action_ section of the test in the `before`.
And in this case, because of the context `'when projects has projects as code owner'`, isn't it executing it twice for those tests?
I'm open to anyone's thoughts about this!

spec/requests/code_owner_projects.rb
hvilloria: JAJAJAJAJAJ the copy pasting doing its work

spec/requests/code_owner_projects.rb
hvilloria: the subject just stores the request in itself but is not executing it, so the it is executed `before` every it in the context that it is declared.

On the other hand if removes the `before` you would have yo put it in every `it` before the `expect`, i think this is just a way to not repeat it every time in the `it`

app/views/shared/user/_sidebar.html.erb
hvilloria: roger that!

PR #228, opened 2020-06-19
Add CodeClimateMetricsUpdaterJob to cron

mrubi-rootstrap:
## What does this PR do?

- It creates the class `CodeClimateMetricsUpdaterJob` to queue the update of CodeClimate metrics
- It adds `CodeClimateMetricsUpdaterJob` to the cron jobs to run it every hour. The reason to run it hourly is that CodeClimate API has an hourly quota on the requests. When the quota is exceeded it will reject requests until the following hour. The cron ensures that if we exceed the quota it will resume where it left on the hour that follows
- It renames and moves the folder of `CodeClimate::UpdateAllProjectsService` to `Processors::CodeClimateUpdateAllProjectsService` to be consistent with the other existing processors


lib/tasks/code_climate.rake
```diff
@@ -1,6 +1,6 @@
 namespace :code_climate do
```
santiagovidal, line 1: Do we still need this rake task?

mrubi-rootstrap: I am not sure. I would keep it in case we want to trigger an update any time other than the cron schedule but I am not sure about it

app/services/processors/code_climate_update_all_projects_service.rb
```diff
@@ -1,5 +1,5 @@
-module CodeClimate
-  class UpdateAllProjectsService < BaseService
+module Processors
+  class CodeClimateUpdateAllProjectsService < BaseService
```
hdamico, line 2: WDYT about creating a folder for CodeClimate services? And then CodeClimate::UpdateAllProjects ?

mrubi-rootstrap: It's already there https://github.com/rootstrap/rs-code-review-metrics/tree/feature/add-code-climate-update-job-cron/app/services/code_climate
but since all the entry points to `Processors` services were in this folder I moved this class here too. I can move it back though

hdamico: Ohh sorry! My bad! Its ok!

mrubi-rootstrap: The thumb would be to leave in `processors` or to move it back to `code_climate` 🤔 ?

lib/tasks/code_climate.rake
santiagovidal: I'm inclined to delete it, to avoid any possible confusions down the road. @santib what's your opinion on this?

mrubi-rootstrap: Deleted it. Thanks

PR #229, opened 2020-06-19
[EM-53] Add global blog metrics

SandroDamilano:
## What does this PR do?
It adds metrics highlights next to the charts in the Blog Metrics tab

image

### DIFF
--- a/app/controllers/tech_blog_controller.rb
+++ b/app/controllers/tech_blog_controller.rb
@@ -20,11 +20,11 @@ class TechBlogController < ApplicationController
   end
 
   def year_to_date_visits
-    Builders::ChartSummary::YearToDateBlogVisits.call.to_i
+    Builders::ChartSummary::YearToDateBlogVisits.call
   end
 
   def month_to_date_visits
-    totals_for(@visits_per_technology)[:data][current_month_key].to_i
+    @visits_per_technology.totals_for(Time.zone.now).to_i
   end
 
   def technology_blog_post_count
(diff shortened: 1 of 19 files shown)


app/controllers/tech_blog_controller.rb
```diff
@@ -11,11 +19,51 @@ def visits_per_technology
     Builders::BlogMetricChart::TechnologyVisits.call
   end
 
-  def visits_growth_mom
-    Builders::BlogMetricChart::TechnologyVisitsGrowthMom.call
+  def year_to_date_visits
+    Builders::ChartSummary::YearToDateBlogVisits.call.to_i
```
mrubi-rootstrap, line 23: Can `YearToDateBlogVisits` return an Integer?
It would uniform it with the other services

app/controllers/tech_blog_controller.rb
```diff
@@ -11,11 +19,51 @@ def visits_per_technology
...
+    totals_for(@visits_growth_mom)[:data][current_month_key].round(2)
+  end
+
+  def last_month_visits_growth
+    totals_for(@visits_growth_mom)[:data][last_month_key].round(2)
+  end
+
+  def totals_for(metric_datasets)
+    metric_datasets.find { |metric_dataset| metric_dataset[:name] == 'Totals' }
+  end
```
mrubi-rootstrap, line 56: This method seems to belong to a BlogMetricsDataset class

It would allow to do

```ruby
@visits_growth_mom.totals[:data][last_month_key].round(2)
```

or even

```ruby
@visits_growth_mom.totals_at(last_month_key)
```

app/controllers/tech_blog_controller.rb
```diff
@@ -11,11 +19,51 @@ def visits_per_technology
...
+  def current_month_key
+    Time.zone.now.strftime(monthly_key_format)
+  end
+
+  def last_month_key
+    Time.zone.now.last_month.strftime(monthly_key_format)
+  end
+
+  def monthly_key_format
+    '%B %Y'
```
mrubi-rootstrap, line 67: This method also seems to belong to a BlogMetricsDataset class.

It would relief the controller from needing to know how to format dates in its internal hash

app/controllers/tech_blog_controller.rb
SandroDamilano: Absolutely yes!

app/controllers/tech_blog_controller.rb
SandroDamilano: Oh, I hadn't thought about that! Love the idea!

spec/controllers/tech_blog_controller_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'rails_helper'
+
+describe TechBlogController, type: :request do
```
santiagovidal, line 3: Given that this is a request spec, perhaps we should change the name of the spec file, and this `describe` as well

PR #230, opened 2020-06-19
[HOTFIX][CodeOwners List]

hvilloria:
## What does this PR do?

* some code owners are not showed because are private
* now authenticates using a github username and access token

ENV VARIABLES:

GITHUB_ADMIN_USER = github_username
GITHUB_ADMIN_TOKEN = personal_access_token

### DIFF
--- a/app/services/github_repository_client.rb
+++ b/app/services/github_repository_client.rb
@@ -1,4 +1,4 @@
-class GithubReposApi
+class GithubRepositoryClient
   URL = 'https://api.github.com/repos/rootstrap'.freeze
 
   def initialize(project_name)
--- a/.env.example
+++ b/.env.example
@@ -4,3 +4,7 @@ GITHUB_ANALYZER_PASSWORD=postgres
 
 # Must be set as the webhook secret on your github app
 GITHUB_ANALYZER_WEBHOOK_SECRET=fake-secret
+
+# must be set to expect the ok process for code owners file job
+GITHUB_ADMIN_USER = github_username
+GITHUB_ADMIN_TOKEN = personal_access_token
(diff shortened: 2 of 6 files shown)


app/services/github_repository_client.rb
```diff
@@ -0,0 +1,15 @@
...
+    @project_name = project_name
+  end
+
+  def get_content_from_file(file_name)
+    connection = Faraday.new(url: "#{URL}/#{@project_name}/contents/#{file_name}") do |conn|
+      conn.basic_auth(ENV['GITHUB_ADMIN_USER'], ENV['GITHUB_ADMIN_TOKEN'])
+    end
+    response = connection.get { |req| req.headers['Accept'] = 'application/vnd.github.v3.raw' }
+    response.success? ? response.body : ''
+  end
```
hdamico, line 14: Shouldn't this be a private method?

app/services/github_repository_client.rb
```diff
@@ -0,0 +1,15 @@
+class GithubRepositoryClient
+  URL = 'https://api.github.com/repos/rootstrap'.freeze
```
hdamico, line 2: Can we extract the organization name to another constant?

app/services/github_repository_client.rb
hvilloria: if this method becomes private we lose the entry point of the class, we don't have a `call` method because the methods of this class should be the different endpoints of the github repos API

app/services/github_repository_client.rb
hvilloria: is there cases where the org could change?

PR #236, opened 2020-06-23
Feature/em 88 add code climate section section to controller

mrubi-rootstrap:
## What does this PR do?

- It adds CodeClimate department information to DevelopmentMetricsController
- It adds CodeClimate summary to Department view


app/services/code_climate/projects_summary.rb
```diff
@@ -21,8 +21,10 @@ def projects_rated_with(letter)
       ratings&.fetch(letter.to_s, 0)
     end
 
-    def rating_names
-      ratings.keys
+    def project_rates(&block)
+      ratings.each_pair(&block)
```
hdamico, line 25: Shouldn't we check if a block is given? Otherwise what would happen here?

mrubi-rootstrap: Yes, the name is misleading since it's an iterator. I will rename it to each_project_rate. Thanks!

PR #238, opened 2020-06-24
Per language metrics

hdamico:
## What does this PR do?

Jun-24-2020 11-47-08

Jira: [EM-64](https://rootstrap.atlassian.net/browse/EM-64)


spec/spec_helper.rb
```diff
@@ -93,4 +93,7 @@
   # as the one that triggered the failure.
   Kernel.srand config.seed
 =end
+  config.before(:suite) do
+    Rails.application.load_seed
```
SandroDamilano, line 97: I don't quite agree with loading the seeds before running the tests. At least not because of the way we are using the seeds, which is in part to populate our local DBs with some data to test our local environment.
If the seeds were being used to populate the production DB in a clean and running state, other would be the case.

SandroDamilano: Never mind, I just realized that the extra data is being created only in development, not test. Sorry for the confusion!

hdamico: Np!

app/models/project.rb
```diff
@@ -45,27 +36,15 @@ class Project < ApplicationRecord
...
-    mobile: %w[ios android react_native].freeze
-  }.freeze
-
-  def associate_department
-    return unless department.nil? || lang_changed?
-
-    department_name = find_department_name(lang)
-    self.department = Department.find_or_create_by!(name: department_name) if department_name
-  end
+  def set_default_language
```
santib, line 45: shouldn't this default be at DB level?

hdamico: A default language_id right? Yep I did not think on that! Thanks

db/migrate/20200622214544_create_languages.rb
```diff
@@ -0,0 +1,8 @@
+class CreateLanguages < ActiveRecord::Migration[6.0]
+  def change
+    create_table :languages do |t|
+      t.string :name
```
santib, line 4: null false?

db/seeds.rb
```diff
@@ -5,54 +5,73 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+unless Rails.env.production?
```
santib, line 8: seeds shouldn't be run in prod so I guess this check is unnecessary?

db/migrate/20200622221335_remove_lang_from_projects.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveLangFromProjects < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :projects, :lang
```
santib, line 3: don't you need this data before dropping the column?

spec/spec_helper.rb
```diff
@@ -93,4 +93,7 @@
   # as the one that triggered the failure.
   Kernel.srand config.seed
 =end
+  config.before(:suite) do
```
santib, line 96: if something happens (e.g. I run ctrl+c) I'll be left out with records in the testing DB, right? (might be worth checking just in case it changed with Rails 6)

If yes, it'd be good to find a way to clear the DB first thing when running tests (even before loading seeds)

spec/spec_helper.rb
santib, line 97: I think a better solution is using fixtures
```
Fixtures is a fancy word for sample data. Fixtures allow you to populate your testing database with predefined data before your tests run. Fixtures are database independent and written in YAML. There is one file per model.
```
https://guides.rubyonrails.org/testing.html#what-are-fixtures-questionmark

db/seeds.rb
hdamico: I put it just in case, but you are right. One less condition

db/migrate/20200622221335_remove_lang_from_projects.rb
hdamico: Not really, we never set it

spec/spec_helper.rb
hdamico: Noice!

PR #241, opened 2020-06-25
[EM-66.2] Add private column to projects

SandroDamilano:
## What does this PR do?
It adds a new column `private` to the projects table, so we can tell which projects are part of Open Source and which are not.

- Added migration
- Created rake task to backfill old projects
- Updated project builder to include the `private` attribute


app/services/processors/project_privacy_backfiller.rb
```diff
@@ -0,0 +1,9 @@
+module Processors
+  class ProjectPrivacyBackfiller < BaseService
+    def call
+      Project.find_each do |project|
+        project.update(is_private: project.events.first&.data&.dig('repository', 'private'))
```
mrubi-rootstrap, line 5: Will `:dig` work even if `first&` or `data&` are nil? `first&` will catch a nil value but I am sure about `nil.dig`

SandroDamilano: I think that's the point in using `&`
If `project.events.first&.data` is `nil`, then `dig('repository', 'private')` will return `nil` as well

spec/services/builders/project_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe Builders::Project do
+  describe '.call' do
+    let(:repository_payload) { create(:repository_payload)['repository'] }
+
+    context 'when the project has not been created' do
+      it 'creates it' do
+        expect { described_class.call(repository_payload) }.to change(Project, :count).by(1)
```
mrubi-rootstrap, line 9: Really not a priority but I would make this assertion in a different test

```ruby
it 'creates it' do
```

```ruby
it 'creates only one' do
```

app/services/processors/project_privacy_backfiller.rb
mrubi-rootstrap: I answer myself, never mind. It won't get called if `:first` is nil.

mrubi-rootstrap:
Yes, I realized that just now.
Thanks!

Martín Rubí
Developer, Rootstrap
Los Angeles / New York / Montevideo

PR #245, opened 2020-06-29
Hotfix per department and language metrics

hdamico:
## What does this PR do?

Add the condition in the query where the metrics that we are retrieving should be created on the same day.

### DIFF
--- a/app/services/metrics/merge_time/per_department.rb
+++ b/app/services/metrics/merge_time/per_department.rb
@@ -13,8 +13,7 @@ module Metrics
 
       def project_metrics_per_department
         Department.joins(languages: { projects: :metrics })
-                  .where(metrics: { name: :merge_time })
-                  .where(metrics: { created_at: metric_interval })
+                  .where(metrics: { name: :merge_time, created_at: metric_interval })
                   .group(:id)
                   .pluck(:id, Arel.sql('COUNT(*), SUM(metrics.value)'))
       end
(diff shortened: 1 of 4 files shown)


app/services/metrics/merge_time/per_department.rb
```diff
@@ -14,6 +14,7 @@ def process
       def project_metrics_per_department
         Department.joins(languages: { projects: :metrics })
                   .where(metrics: { name: :merge_time })
+                  .where(metrics: { created_at: metric_interval })
```
horacio, line 17:
```suggestion
                  .where(metrics: { name: :merge_time, created_at: metric_interval })
```

app/services/metrics/merge_time/per_language.rb
```diff
@@ -14,6 +14,7 @@ def process
       def project_metrics_per_language
         Language.joins(projects: :metrics)
                 .where(metrics: { name: :merge_time })
+                .where(metrics: { created_at: metric_interval })
```
horacio, line 17: Same as above

app/services/metrics/review_turnaround/per_department.rb
```diff
@@ -14,6 +14,7 @@ def process
       def project_metrics_per_department
         Department.joins(languages: { projects: :metrics })
                   .where(metrics: { name: :review_turnaround })
+                  .where(metrics: { created_at: metric_interval })
```
horacio, line 17: Same as above

app/services/metrics/review_turnaround/per_language.rb
```diff
@@ -14,6 +14,7 @@ def process
       def project_metrics_per_language
         Language.joins(projects: :metrics)
                 .where(metrics: { name: :review_turnaround })
+                .where(metrics: { created_at: metric_interval })
```
horacio, line 17: Same as above

PR #247, opened 2020-07-01
[EM-66.3] Create open source visits metrics

SandroDamilano:
## What does this PR do?
It creates the `open_source_visits` metric for projects.
It includes:
- `OpenSourceProjectViewsUpdater` processor to update every project repository views on the `Metrics` table
- Schedule the job to update all Open Source metrics once per day

### DIFF
--- a/app/services/processors/open_source_project_views_updater.rb
+++ b/app/services/processors/open_source_project_views_updater.rb
@@ -1,28 +1,29 @@
 module Processors
   class OpenSourceProjectViewsUpdater < BaseService
+    def initialize(project)
+      @project = project
+    end
+
     def call
-      Project.open_source.each do |project|
-        views_to_update_for(project).each do |project_views_payload|
-          update_views_metric_for_project(
-            project,
-            project_views_payload['timestamp'],
-            project_views_payload['uniques']
-          )
-        end
+      views_to_update.each do |project_views_payload|
+        update_views_metric(
+          project_views_payload['timestamp'],
+          project_views_payload['uniques']
+        )
       end
     end
 
     private
 
-    def views_to_update_for(project)
-      last_updated_timestamp = last_updated_timestamp_for(project)
+    attr_reader :project
 
-      views_for_project(project)['views'].select do |views_payload|
+    def views_to_update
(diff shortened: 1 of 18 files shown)


app/services/processors/open_source_project_views_updater.rb
```diff
@@ -0,0 +1,48 @@
+module Processors
+  class OpenSourceProjectViewsUpdater < BaseService
+    def call
+      Project.open_source.each do |project|
```
santib, line 4: not sure if it's possible, but if you move this loop to `update_open_source_visits`, you can have `OpenSourceProjectViewsUpdater` work with 1 project at a time, and you don't need to pass around the project as an argument. Feel free to ignore this comment

app/services/processors/open_source_project_views_updater.rb
```diff
@@ -0,0 +1,48 @@
...
+        ownable: project
+      ).maximum(:value_timestamp) || Time.zone.at(0)
+    end
+
+    def views_for_project(project)
+      GithubRepositoryClient.new(project.name).repository_views
+    end
+
+    def update_views_metric_for_project(project, timestamp, views)
+      metric = Metric.find_or_initialize_by(
```
santib, line 38: just to confirm, this is going to update outdated metrics, right? e.g. I run this today and I get 5 visits in my project for today, but there are still more more hours in this day, so if I run it tomorrow it will update it to say 20 visits.

SandroDamilano: Yes! That's what it will do. It's the same approach I took with the blog visit metrics.

app/services/processors/open_source_project_views_updater.rb
SandroDamilano: It's funny, because when I finished coding this class that's exactly what I thought, that I could have given this the responsibility of updating just one project and not all of them.
I feel like if we changed it, we would be able to relieve this class of the responsibility of having to know which projects it has to update, which seems like a good move.
So no, I'm not ignoring this comment, I'll make the change 😂

PR #248, opened 2020-07-01
Review turnaround table

hdamico:
## What does this PR do?

Creates Review turnaround table, which is going to store every turnaround value from each review.


app/models/events/review.rb
```diff
@@ -46,5 +46,15 @@ class Review < ApplicationRecord
 
     validates :state, inclusion: { in: states.keys }
     validates :github_id, :opened_at, presence: true
+
+    after_create :build_review_turnaround
```
santib, line 50: what happens with existing records? are we going to backfill or ignore them?

app/models/review_turnaround.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  private
+
+  def build
+    self.value = calculate_turnaround
+    save!
+  end
+
+  def calculate_turnaround
+    review_request.reviews.first.opened_at.to_i - review_request.created_at.to_i
```
santib, line 29: so are you sure `review_request.reviews.first` will always return the oldest review? is the `order` specified somewhere?

app/models/events/review.rb
santib, line 50: maybe you can add a test for this business logic?

hdamico: I could build a rake for that. WDYT?

app/models/review_turnaround.rb
santiagovidal: Perhaps this record should hold a reference to the `review`

app/models/events/review.rb
santib: Sounds good to me

app/models/review_turnaround.rb
hdamico: We have the condition in Review model where we ask if the review request has one review.

app/models/events/review.rb
hdamico: Yes! I forgot

app/models/review_turnaround.rb
```diff
@@ -0,0 +1,34 @@
...
+#  fk_rails_...  (review_request_id => review_requests.id)
+#
+
+class ReviewTurnaround < ApplicationRecord
+  belongs_to :review_request
+
+  validates :value, presence: true
+  validates :review_request_id, uniqueness: true
+
+  def set_value
```
santib, line 24: minor comment, but this business logic code could be moved to a service object so to keep models thin

spec/models/events/review_spec.rb
```diff
@@ -51,4 +51,28 @@
       expect(subject).to_not be_valid
     end
   end
+
+  describe 'callbacks' do
+    context 'when a review is created' do
```
santib, line 56: shouldn't this be triggered `when the second review is created` ?

spec/models/events/review_spec.rb
```diff
@@ -51,4 +51,28 @@
...
+      let(:review) { create :review, review_request: review_request }
+      let(:review_request) { create(:review_request) }
+
+      it 'triggers review turnaround creation' do
+        expect_any_instance_of(ReviewTurnaround).to receive_message_chain(:set_value, :save!)
```
santib, line 61: this test is just testing that a method is called, it currently works but IMHO it's heavily tied to implementations details. I think a test like the following will be actually checking what you really care about

```ruby
it 'creates a review turnaround with the correct values' do
  expect { review }.to change { ReviewTurnaround.count }.from(0).to(1)

  review_turnaround = ReviewTurnaround.last
  expect(review_turnaround.[some_attribute]) = some_expected_value
  ...
  expect(review_turnaround.[some_attribute]) = some_expected_value
end
```

app/models/review_turnaround.rb
hdamico: Model validations too? Or just the set_value method?

santib: just the `set_value`, e.g. instead of calling `set_value` from the other model you could call `ReviewTurnaroundCreator.new(self).call` . what do you think?

PR #249, opened 2020-07-03
Department distribution

hdamico:
## What does this PR do?

Creates the workflow for department distribution.
Retrieves all the review turnarounds which has a review request from a project which has a language of that department, then retrieves their values and group them the intervals 1-12, 12-24, 24-36, 36-48, 48-60, 60-72 and more than 72 (72+).


.rubocop.yml
```diff
@@ -46,7 +46,7 @@ Metrics/ClassLength:
 
 # Avoid complex methods.
 Metrics/CyclomaticComplexity:
-  Max: 6
+  Max: 7
```
santiagovidal, line 49: Wouldn't it make more sense to ignore a specific method instead of increasing the threshold? Which method caused this?

app/services/builders/chartkick/base.rb
```diff
@@ -9,15 +9,43 @@ def initialize(entity_id, query)
...
+        end
+      end
+
+      def value_in_hours_for(entity)
+        (entity.value.to_f / 1.hour.seconds).round(2)
+      end
+
+      def resolve_interval(entity)
+        entity_value = entity.value
+        if entity_value < 12
```
santiagovidal, line 34: Perhaps we can have a distributions hash, like:

```
{
  0..12 => '1-12',
  12..24 => '12-24'
...
}
```

And when you have a value, you check which of the hash keys contains your value. I don't know if it simplifies the code, but at least it will make rubocop and codeclimate stop complaining.

hdamico:
```
                    user        system         total         real
without hash      0.001478     0.000003       0.001481    (0.001478)
with hash         0.064149     0.000294       0.064443    (0.064776)
```
The code for `with hash` 
```
INTERVALS.select { |interval, _| interval.include?(entity_value) }.values.first
```

santiagovidal: I'm fine with either approach 👍

santib, line 34: what about using a `case` instead of if elsif?

app/services/builders/chartkick/department_distribution_data.rb
```diff
@@ -0,0 +1,20 @@
...
+        review_turnarounds = retrieve_review_turnarounds
+        [{ name: department_name, data: build_distribution_data(review_turnarounds) }]
+      end
+
+      private
+
+      def retrieve_review_turnarounds
+        ::ReviewTurnaround.joins(review_request: { project: { language: :department } })
+                          .where(departments: { id: @entity_id })
+                          .where(created_at: @query[:value_timestamp])
```
santib, line 16: what is `@query[:value_timestamp]` ??

app/services/builders/chartkick/base.rb
hdamico: Tried it, but less performing.

db/migrate/20200701133311_create_review_turnarounds.rb
```diff
@@ -3,6 +3,8 @@ def change
     create_table :review_turnarounds do |t|
       t.references :review_request, null: false, foreign_key: true
       t.integer :value
+
+      t.timestamps
```
santib, line 7: you shouldn't be editing existing migrations (if they were already run in Heroku), you need to create new migrations

hdamico: Is not in heroku yet

app/services/builders/chartkick/department_distribution_data.rb
hdamico: I had to adapt the code to the old one. @query is a method in Base
```
      #app/services/metric/group/base.rb
      def query
        {
          name: @metric_name,
          value_timestamp: value_timestamp,
          interval: interval
        }
      end

      #app/services/metric/group/weekly.rb
      def value_timestamp
        (current_time - @prev.weeks).beginning_of_week..current_time.end_of_week
      end
```

PR #250, opened 2020-07-06
[EM-66.4] Show open source visits metrics

SandroDamilano:
## What does this PR do?
It adds the logic to build and show the metrics for Open Source unique visits.

It also changes the structure for the blog metrics builders, so that the code can be reused for the open source ones.

Before:
```
builders
  |_ blog_metric_chart
        |_ base.rb
        |_ technology_visits.rb
        ...
```
Now:
```
builders
  |_ metric_chart
        |_ blog
        |     |_ base.rb
        |     |_ technology_visits.rb
        |     ...
        |_ open_source
        |     |_ base.rb
        |     |_ language_visits.rb
        |_ base.rb
```


app/services/builders/metric_chart/base.rb
```diff
@@ -1,26 +1,26 @@
...
         @months = months
       end
 
       def call
-        datasets = Technology.all.map do |technology|
+        datasets = entity_type.all.map do |entity|
           generate_results_for(
-            entity_name: technology.name.titlecase,
-            metrics: technology.metrics,
+            entity_name: entity_name(entity),
```
santib, line 11: when a base class requires the actual implementations to define certain methods, I like to make it explicit like

```ruby
        def entity_name(_technology)
          raise NotImplementedError
        end
```

But I know my friend @juanmanuelramallo don't like this pattern. So feel free to ignore it 😄

SandroDamilano: It's funny because I used to do what you say and then we discussed this with him [here](https://github.com/rootstrap/best_buy_ruby/pull/8#discussion_r393093188).
I do agree with keeping it documented in some way. Even better if it's in a _runnable_ way (not just comments).
Maybe we can create some sort of custom exception for these cases? Or we could just bring the discussion to #ruby and fight to death (?

juanmanuelramallo: hahaha I no longer have a strong opinion on this. As long as we follow this approach everywhere in the codebase I'm ok with it 👌

app/services/builders/metric_chart/blog/growth_month_over_month.rb
```diff
@@ -0,0 +1,26 @@
...
+      class GrowthMonthOverMonth < Builders::MetricChart::Blog::Base
+        def initialize(months = 13)
+          super(months + 1)
+        end
+
+        private
+
+        def process_data(metrics_data)
+          metrics_data.each_with_object({}) do |(month, metric_value), hash|
+            previous_metric_value = metrics_data[month.last_month]
```
santib, line 13: just to check: is `month.last_month` the previous month of the one in the iteration?

app/services/builders/metric_chart/blog/growth_month_over_month.rb
```diff
@@ -0,0 +1,26 @@
...
+            next(hash) if previous_metric_value.nil? || previous_metric_value.zero?
```
santib, line 14:
```suggestion
            next(hash) unless previous_metric_value&.positive?
```

minor comment

app/services/builders/metric_chart/blog/growth_month_over_month.rb
```diff
@@ -0,0 +1,26 @@
...
+
+            hash[month] = growth_for(metric_value, previous_metric_value)
+          end
+        end
+
+        def growth_for(metric_value, previous_metric_value)
+          (metric_value - previous_metric_value) / previous_metric_value * 100
```
santib, line 21: I think this calc is wrong

```
[4] pry(main)> (3 - 2) / 2 * 100
=> 0
[5] pry(main)> (3 - 2) * 100 / 2
=> 50
```

app/services/builders/metric_chart/blog/growth_month_over_month.rb
SandroDamilano: Correct!

app/services/builders/metric_chart/blog/growth_month_over_month.rb
SandroDamilano: I hadn't thought of that! Thanks!

app/services/builders/metric_chart/blog/growth_month_over_month.rb
SandroDamilano: Great finding! I think it still works alright because the metric values extracted from the `metrics` table are floats, not integers. So the calculation would actually be like this:
```
[1] pry(main)> (3.0 - 2.0) / 2.0 * 100
 => 50.0 
```
Still, I'm gonna change it just in case

santib: oh good point 👍

spec/services/builders/metric_chart/base_spec.rb
```diff
@@ -62,4 +62,28 @@
       end
     end
   end
+
+  describe 'abstract methods' do
+    subject { described_class.send(:new) }
```
santib, line 67:
```suggestion
    subject { described_class.new }
```

SandroDamilano: `new` is a private method for all classes in the `BaseService` hierarchy. That's why we can't do `described_class.new` directly.

PR #251, opened 2020-07-08
[EM-66.5] Make open source visits metrics weekly

SandroDamilano:
## What does this PR do?
It changes the way we calculate and show Open Source metrics. We used to fetch the daily visits and show them grouped by month. But as we're looking for unique visits, adding each month daily unique visits does not equal to the monthly unique visits.
So given that the biggest period the Github API returns the unique visits is weekly, we'll now get and show them weekly.

image

### DIFF
--- a/app/services/builders/metric_chart/base.rb
+++ b/app/services/builders/metric_chart/base.rb
@@ -1,12 +1,12 @@
 module Builders
   module MetricChart
     class Base < BaseService
-      def initialize(months = 13)
-        @months = months
+      def initialize(periods = 13)
+        @periods = periods
       end
 
       def call
-        datasets = entity_type.all.map do |entity|
+        datasets = entities.map do |entity|
           generate_results_for(
             entity_name: entity_name(entity),
             metrics: entity_metrics(entity),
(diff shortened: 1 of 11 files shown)


app/services/builders/metric_chart/base.rb
```diff
@@ -70,6 +70,18 @@ def metric_ownable_type
...
+        raise NoMethodError
+      end
+
+      def chart_date_format
+        raise NoMethodError
+      end
+
+      def grouping_period
+        raise NoMethodError
+      end
```
hdamico, line 84: I'm wondering which is the difference of not having this, if you call a method that is not implemented you are going to get a `NoMethodError` anyway, don't you?

SandroDamilano: Yes, you're totally correct. What we discussed is the necessity of having some live documentation in our code. Defining these methods like this is surely the same as not doing so from an execution standpoint, but making it explicit like this makes the devs understand better which are the methods we should be implementing in the subclasses.

colembolos: I had some discussions about this in other projects. Why are we trying to implement some interfaces like other languages ​​(like Java)? I think it's an additional layer that we don't need, I know the meaning is documenting, but for ruby on rails ​​developers it is not necessary in my opinion

SandroDamilano: Alright, given that none of us is sure about this approach and that the rest of the code is not written this way, I'm just gonna delete all these abstract methods.
But I'll do that in other PR.

PR #256, opened 2020-07-13
Department merge time distribution

hdamico:
## What does this PR do?

Creates department distribution for the merge time

### DIFF
--- a/app/models/events/pull_request.rb
+++ b/app/models/events/pull_request.rb
@@ -58,8 +58,12 @@ module Events
               inclusion: { in: [true, false] }
     validates :github_id, uniqueness: true
 
-    ##
-    # Returns the number of reviews this pull_request has
-    delegate :count, to: :reviews, prefix: true
+    after_validation :build_merge_time, on: :update, if: :merged_at_changed?
+
+    private
+
+    def build_merge_time
+      Builders::MergeTime.call(self)
+    end
   end
 end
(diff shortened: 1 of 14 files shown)


app/models/events/pull_request.rb
```diff
@@ -58,8 +58,12 @@ class PullRequest < ApplicationRecord
               inclusion: { in: [true, false] }
     validates :github_id, uniqueness: true
 
-    ##
-    # Returns the number of reviews this pull_request has
-    delegate :count, to: :reviews, prefix: true
+    after_validation :build_merge_time, on: :update, if: :merged_at_changed?
```
santiagovidal, line 61: Perhaps the following is a bit simpler:

`after_update :build_merge_time, if: :merged_at_changed?`

WDYT?

spec/models/events/pull_request_spec.rb
```diff
@@ -62,4 +62,24 @@
...
+
+  describe 'callbacks' do
+    context 'when a pull request is merged' do
+      before { travel_to(Time.zone.today.beginning_of_day) }
+      let!(:pull_request) { create(:pull_request, opened_at: Time.current) }
+
+      it 'creates a merge time with the correct values' do
+        pull_request.update!(merged_at: Time.current)
+        merge_time = MergeTime.last
+        expect(merge_time[:value]).to eq(0)
```
santiagovidal, line 74: Using `0` is a bit confusing to me. Perhaps you can do:

`pull_request.update!(merged_at: Time.current + 1.hour)`

and then

`expect(merge_time[:value]).to eq(3600)`

WDYT?

app/models/events/pull_request.rb
hdamico: After_update is obsolete https://apidock.com/rails/v2.3.8/ActiveRecord/Callbacks/after_update.

PR #260, opened 2020-07-16
Fix Review Turnaround generation

hdamico:
## What does this PR do?

Fixes the misunderstanding of review turnaround. We were calculating the review turnaround as the first review of a user, which is incorrect, the correct way is to calculate it as the first review of a pull request.

### DIFF
--- a/app/models/review_turnaround.rb
+++ b/app/models/review_turnaround.rb
@@ -2,24 +2,24 @@
 #
 # Table name: review_turnarounds
 #
-#  id                :bigint           not null, primary key
-#  value             :integer
-#  created_at        :datetime         not null
-#  updated_at        :datetime         not null
-#  review_request_id :bigint           not null
+#  id              :bigint           not null, primary key
+#  value           :integer
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  pull_request_id :bigint
 #
 # Indexes
 #
-#  index_review_turnarounds_on_review_request_id  (review_request_id)
+#  index_review_turnarounds_on_pull_request_id  (pull_request_id)
 #
 # Foreign Keys
 #
-#  fk_rails_...  (review_request_id => review_requests.id)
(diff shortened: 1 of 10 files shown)


app/models/review_turnaround.rb
```diff
@@ -2,24 +2,24 @@
...
 #
 # Foreign Keys
 #
-#  fk_rails_...  (review_request_id => review_requests.id)
+#  fk_rails_...  (pull_request_id => pull_requests.id)
 #
 
 class ReviewTurnaround < ApplicationRecord
-  belongs_to :review_request
+  belongs_to :pull_request, class_name: 'Events::PullRequest'
```
santiagovidal, line 21: If we associate with the PR, will we have problems if we want to have different metrics in the future? For example if we want to track `Time to first review`?

app/services/builders/review_turnaround.rb
```diff
@@ -1,17 +1,17 @@
...
     def call
-      ::ReviewTurnaround.create!(review_request: @review_request, value: calculate_turnaround)
+      ::ReviewTurnaround.create!(pull_request: @pull_request, value: calculate_turnaround)
     end
 
     private
 
     def calculate_turnaround
-      @review_request.reviews.first.opened_at.to_i - @review_request.created_at.to_i
+      @pull_request.opened_at.to_i - @pull_request.reviews.first.created_at.to_i
```
santiagovidal, line 14: This should be the first review by the second reviewer, right?

db/structure.sql
```diff
@@ -1645,6 +1645,6 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20200625144922'),
 ('20200630165139'),
 ('20200701133311'),
-('20200703141617');
-
-
+('20200703141617'),
+('20200713152004'),
```
SandroDamilano, line 1649: Where is this migration?

PR #263, opened 2020-07-20
[EM-109] Support multiple technologies per blog post

SandroDamilano:
## What does this PR do?
It allows blog posts to have more than one technology associated.

To do so, it:
- Creates the `BlogPostTechnology` model, which corresponds to the new `blog_post_technologies` joined table between blog_posts and technologies
- Updates the `BlogPostsUpdater` to add every matching technology to the blog post
- Changes the way metric totals are calculated. As calculating them through the technology metrics may result in counting the same blog post metric more than once, they are now calculated directly from the blog posts metrics.

**Note**
To prevent any information loss, this PR **will not** get rid of the `technology_id` column in `blog_posts` (though it's already not used anymore in this PR). Once we confirm the new implementation is working correctly, I'll create another PR to get rid of it.

### DIFF
--- a/app/services/builders/chart_summary/year_to_date_blog_visits.rb
+++ b/app/services/builders/chart_summary/year_to_date_blog_visits.rb
@@ -3,8 +3,8 @@ module Builders
     class YearToDateBlogVisits < BaseService
       def call
         Metric
-          .where(name: Metric.names[:blog_visits], ownable_type: Technology.to_s)
-          .where('value_timestamp > ?', Time.zone.now.beginning_of_year)
+          .where(name: Metric.names[:blog_visits], ownable_type: BlogPost.to_s)
+          .where('value_timestamp >= ?', Time.zone.now.beginning_of_year)
           .sum(:value).to_i
       end
     end
--- a/spec/models/blog_post_technology_spec.rb
+++ b/spec/models/blog_post_technology_spec.rb
@@ -1,0 +1,25 @@
+# == Schema Information
+#
+# Table name: blog_post_technologies
+#
+#  id            :bigint           not null, primary key
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#  blog_post_id  :bigint           not null
+#  technology_id :bigint           not null
+#
+# Indexes
+#
+#  index_blog_post_technologies_on_blog_post_id   (blog_post_id)
+#  index_blog_post_technologies_on_technology_id  (technology_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (blog_post_id => blog_posts.id)
+#  fk_rails_...  (technology_id => technologies.id)
+#
+
+require 'rails_helper'
+
+RSpec.describe BlogPostTechnology, type: :model do
+end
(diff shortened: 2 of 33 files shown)


app/services/builders/chart_summary/year_to_date_blog_visits.rb
```diff
@@ -3,8 +3,8 @@ module ChartSummary
     class YearToDateBlogVisits < BaseService
       def call
         Metric
-          .where(name: Metric.names[:blog_visits], ownable_type: Technology.to_s)
-          .where('value_timestamp > ?', Time.zone.now.beginning_of_year)
+          .where(name: Metric.names[:blog_visits], ownable_type: BlogPost.to_s)
```
santib, line 6: just curious, but what is this change doing in terms of UI? 

Do we need to reconcile all existing data so that the ownable is now the blog post and not the technology?

spec/models/blog_post_technology_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (blog_post_id => blog_posts.id)
+#  fk_rails_...  (technology_id => technologies.id)
+#
+
+require 'rails_helper'
+
+RSpec.describe BlogPostTechnology, type: :model do
+end
```
santib, line 25: 👀

app/services/builders/chart_summary/year_to_date_blog_visits.rb
SandroDamilano: The problem is that now that a single blog post may belong to more than one technology, its number of visits may be duplicated across those technologies.
For instance, imagine having a single blog post, with both `ruby` and `react` technologies. If that blog post receives 100 visits in any month, both `ruby` and `react` will have 100 visits each. So when I want to get the total number of visits across all technologies, if I just add the visits for each technology I'll end up with 200, which is false. That's why I need to add the data directly from the blog posts visits.

spec/models/blog_post_technology_spec.rb
SandroDamilano: Oh, this is the auto generated spec file created for the new model. I didn't think of any meaningful spec I could add, so I left it blank.
I could add one just for the sake of not leaving it empty, if you'd prefer it that way. I wouldn't mind at all.

santib: I'm ok leaving it empty

app/services/builders/chart_summary/year_to_date_blog_visits.rb
santib: Makes sense. And do we need to update existing records? As far as I understand from this code, metrics used to have the technology as the ownable, and now they'll have the blog post, right?

SandroDamilano: The best part is that the visits metrics for blog posts are already there. We create visits metrics for both blog posts and technologies so there's no need to update any records because of that 😎 
But of course, we will update the blog posts associated technologies, so the metrics for technologies will have to be updated as well. It's not a big problem, though, we just need to run 2 jobs (one that updates the blog posts and another that updates the metrics). Both jobs are already created, so no further PRs are needed.

PR #264, opened 2020-07-21
[feature][Second Review]

hvilloria:
## What does this PR do?

* implements second review for a pull request
* uses that model to show the distribution department data

[Jira ticket EM-116](https://rootstrap.atlassian.net/browse/EM-116)


app/models/events/review.rb
```diff
@@ -56,5 +56,15 @@ def build_review_turnaround
...
     end
+
+    def build_second_review_turnaround
+      return unless second_review_with_different_users?
+
+      Builders::SecondReviewTurnaround.call(self)
+    end
+
+    def second_review_with_different_users?
+      pull_request.reviews.select(:owner_id).distinct.count == 2
```
hvilloria, line 67: @santiagovidal @santib @SandroDamilano guys, i am not sure about this query but
`Events::Review.joins(:pull_request).where(pull_request_id: self.pull_request_id).select(:owner_id).distinct.count == 2` this one i think it is not right because emulation of a class method right?

app/models/second_review_turnaround.rb
```diff
@@ -0,0 +1,25 @@
...
+# Indexes
+#
+#  index_second_review_turnarounds_on_review_request_id  (review_request_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (review_request_id => review_requests.id)
+#
+
+class SecondReviewTurnaround < ApplicationRecord
```
hvilloria, line 20: @SandroDamilano @santiagovidal @santib  wdyt about the name of the table/model? i couldn't imagine any better 😅

app/models/events/review.rb
SandroDamilano: I may be missing something but I don't see anything wrong with any of both 🤔 

I do want to point out that I think we need another condition, because a `SecondReviewTurnaround` will be created every time a new review is submitted by any of the two original reviewers.
Let's say someone creates a PR and I review it first. I give it my approve.
Then you review it but request some changes. A `SecondReviewTurnaround` is created.
The comments are addressed and you review it again, approving it. Another `SecondReviewTurnaround` will be created, because there have been still 2 reviewers so far, you and me.
We probably need to add a condition that checks that the reviewer has not made any other review to this PR. That should be enough.

app/models/second_review_turnaround.rb
SandroDamilano: It doesn't sound bad. It represents the review turnaround of every second review after all, doesn't it?

spec/services/builders/review_turnaround_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+        review_turnaround = ReviewTurnaround.last
+        expect(review_turnaround[:value]).to eq(0)
+        expect(review_turnaround[:review_request_id]).to eq(review_request.id)
+      end
+
+      it 'creates a review turnaround' do
+        expect { review }.to change { ReviewTurnaround.count }.from(0).to(1)
+      end
+
+      context 'when there is more than one review in a review request' do
```
santiagovidal, line 21: I feel like it's the responsibility of the model to test this, not the builder.

spec/models/events/review_spec.rb
```diff
@@ -52,29 +52,41 @@
...
-        it 'does not create review turnaround' do
-          expect { review }.to_not change { ReviewTurnaround.count }
-        end
+    let(:second_rr) do
+      create(:review_request, owner: vita, reviewer: hernan, project: project, pull_request: pr)
+    end
+    let(:second_review) do
+      create(:review, owner: hernan, project: project, pull_request: pr, review_request: second_rr)
+    end
+    context 'only when is the second review for different users' do
```
santiagovidal, line 85: Can we add the opposite context, to make sure Builders::SecondReviewTurnaround is not called when it shouldn't be called? Same with the above context.

app/models/second_review_turnaround.rb
santib: I'm ok with it, but maybe the actual entity is not just the second review, but the "last awaited review"? Like, the last of the reviews that were required to be able to merge it (best case scenario when there are no comments). I'm thinking that if tomorrow there are projects that require different amounts of reviews this modeling won't suffice the requirements.

That said, I'm ok leaving it like this. Chances are that we won't need to do anything like that 😄

app/services/builders/second_review_turnaround.rb
```diff
@@ -0,0 +1,18 @@
+module Builders
+  class SecondReviewTurnaround < BaseService
+    def initialize(review)
+      @review = review
+    end
+
+    def call
+      ::SecondReviewTurnaround.create!(
+        review_request_id: @review.review_request_id,
```
SandroDamilano, line 9: Could the review not have an associated review request? What would we do in that case?

hvilloria: it couldn't, it is a fk,  it will raise an exception

SandroDamilano: Alright, but we should handle that exception, shouldn't we? Because we're not preventing this builder to be called with a review without review request.

app/models/completed_review_turnaround.rb
```diff
@@ -0,0 +1,25 @@
...
+#
+#  index_completed_review_turnarounds_on_review_request_id  (review_request_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (review_request_id => review_requests.id)
+#
+
+class CompletedReviewTurnaround < ApplicationRecord
+  belongs_to :review_request
```
santib, line 21: just curious, but why do we need the reference to just the last one and not to all the reviews that made this record to be created?

spec/models/events/review_spec.rb
```diff
@@ -53,27 +53,85 @@
...
+      let(:third_review) do
+        create(:review,
+               owner: hernan, project: project, pull_request: pr, review_request: second_rr)
+      end
+
+      context 'when a different user make the second review' do
+        it 'builds the complete review turnaround' do
+          expect(Builders::CompletedReviewTurnaround)
+            .to receive(:call)
+          second_review
```
santib, line 124: not necessary to change it now, but I think it'd be more valuable a test like this
```ruby
expect { second_review }.to change(CompletedReviewTurnaround).by(1)
```

because it's testing behavior instead of implementation

app/models/completed_review_turnaround.rb
santib, line 21: why are we storing the reference to only one review request?

hvilloria: i think this was the simplest association because we can navigate through all the info from here, it would be harder to collect information about the review if we store the reference of a PR or something like that

app/services/builders/second_review_turnaround.rb
SandroDamilano: So no, basically we don't have any review without a review request. We are safe there.

spec/models/events/review_spec.rb
hvilloria: i will change it anyway, it has more sense.

PR #273, opened 2020-07-24
[BUG][Code Owners File]

hvilloria:
## What does this PR do?

* File content is requested in three different locations

[Jira bug EM-126](https://rootstrap.atlassian.net/browse/EM-126)

### DIFF
--- a/app/services/github_repository_client.rb
+++ b/app/services/github_repository_client.rb
@@ -20,7 +20,7 @@ class GithubRepositoryClient
   private
 
   def find_in_locations
-    content = LOCATIONS.each do |location|
+    LOCATIONS.each do |location|
       connection = Faraday.new(url: "#{URL}/#{@project_name}/contents/#{location}") do |conn|
         conn.basic_auth(ENV['GITHUB_ADMIN_USER'], ENV['GITHUB_ADMIN_TOKEN'])
       end
@@ -28,6 +28,6 @@ class GithubRepositoryClient
       response = connection.get { |req| req.headers['Accept'] = 'application/vnd.github.v3.raw' }
       return response.body if response.success?
     end
-    content.is_a?(Array) ? '' : content
+    ''
   end
 end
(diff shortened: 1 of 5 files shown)


app/services/github_repository_client.rb
```diff
@@ -18,4 +17,16 @@ def repository_views
     response = connection.get { |request| request.params['per'] = 'week' }
     JSON.parse(response.body).with_indifferent_access
   end
+
+  private
+
+  def find_in_locations
+    LOCATIONS.map do |location|
```
santiagovidal, line 24: Instead of a `map`, perhaps we can do `each` and call `break` once we find the file, so that we don't perform unnecessary API calls.

hvilloria: to be honest i didn't do the break because i thought it was not a rails way 😆 , i will change it inmediately 👍

santiagovidal: `detect` might work, too

app/services/github_repository_client.rb
```diff
@@ -18,4 +16,18 @@ def repository_views
...
+  def find_in_locations
+    content = LOCATIONS.each do |location|
+      connection = Faraday.new(url: "#{URL}/#{@project_name}/contents/#{location}") do |conn|
+        conn.basic_auth(ENV['GITHUB_ADMIN_USER'], ENV['GITHUB_ADMIN_TOKEN'])
+      end
+
+      response = connection.get { |req| req.headers['Accept'] = 'application/vnd.github.v3.raw' }
+      return response.body if response.success?
+    end
+    content.is_a?(Array) ? '' : content
```
SandroDamilano, line 31: What is this line supposed to do?

hvilloria: if the array doesn't return anything, stores the initial array of locations in the variable so when that happens we want to return an empty string instead of the array of locations

SandroDamilano: In that case, why are we checking what's in `content`? If we made it to this line, then the file couldn't be found in any location. In which case we should just return `''`
Does it make sense?

PR #276, opened 2020-07-29
[EM-130] Update CodeClimate summary for department metrics

SandroDamilano:
## What does this PR do?
It updates the view of the CoceClimate summary for department metrics

image

Resolves [#130](https://rootstrap.atlassian.net/browse/EM-130)


app/assets/stylesheets/metrics/_code-climate-deparments.scss
```diff
@@ -27,3 +27,30 @@ table.code-climate-projects-report {
...
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+}
+
+.code-climate-summary-issues {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+}
```
hvilloria, line 56: maybe in the future but since bootstrap has some utils we could use them, i can not remember exactly how but it is something like `class='d-flex justify-content'`

SandroDamilano: Yeah, I've seen that used in the code I replaced. Do we prefer using those classes instead of creating ours and setting the styles in the css files?
It's just my opinion (and not a strong one, for sure, so I'd love to hear your thoughts) that creating our own classes may add semantics to the code, and that using those classes does not sound very different than adding the styles inline in the html (like `style='display: flex; justify-content: space-between;'`)

hdamico: I like to have new css classes with more sense than the ones that bootstrap already has, but maybe we can reduce the amount, wdyt? For example .code-climate-summary-ratings and .code-climate-summary-issues are the same, could we use something more general in this case to wrap both?

hdamico: Another way to wrap all these classes could be:
```
.code-climate-summary {
  # general styles...
  .container { ... }
  .title { ... }
  .info { ... }
}
```

SandroDamilano: I like that alternative! And thinking about both classes you mentioned, I could join them in something like `code-climate-summary-item-list`, since that's basically what they have in common.
Thanks!

PR #277, opened 2020-07-30
[feature] App design changed

hvilloria:
## What does this PR do?

Screen Shot 2020-07-30 at 2 38 51 PM


app/views/development_metrics/departments.erb
```diff
@@ -1,42 +1,51 @@
...
+                <div class="col-md-7">
+                </div>
+              </div>
+            <% end %>
           </div>
-          <div class="col-md-7">
+        </div>
+        <% if @review_turnaround && @merge_time %>
+          <div class="row">
+            <%= render 'development_metrics/code_climate/code_climate_department_summary',
```
santiagovidal, line 19: Shouldn't we render this even if there is no `@review_turnaround` and `@merge_time`?

hvilloria: uhmm 🤔 i think that would be different based on the design

santiagovidal: Would it? If there is no RT or MT data, but there is CC data, shouldn't we show the CC data?

PR #278, opened 2020-07-31
[EM-132] Add relevance to projects

SandroDamilano:
## What does this PR do?
It adds the `relevance` attribute to projects, which can be either:
- `commercial`
- `internal`
- `ignored`
- `unassigned`

Now, Open Source projects are those that meet all these conditions:
- Are not private (`is_private` set to `false`) (same as before)
- Have a language (`language` is not `nil` nor `unassigned`) (same as before)
- Are Internal (`relevance` is `internal`) (new condition)

Resolves [132](https://rootstrap.atlassian.net/browse/EM-132)

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -6,6 +6,7 @@
 #  description :string
 #  is_private  :boolean
 #  name        :string
+#  relevance   :enum             default("unassigned"), not null
 #  created_at  :datetime         not null
 #  updated_at  :datetime         not null
 #  github_id   :integer          not null
@@ -17,6 +18,13 @@
 #
 
 class Project < ApplicationRecord
+  enum relevance: {
+    commercial: 'commercial',
+    internal: 'internal',
+    ignored: 'ignored',
+    unassigned: 'unassigned'
+  }
+
   belongs_to :language
 
   has_many :events, dependent: :destroy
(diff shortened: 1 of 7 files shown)


app/models/project.rb
```diff
@@ -38,17 +46,22 @@ class Project < ApplicationRecord
            source: :user
 
   validates :github_id, presence: true, uniqueness: true
+  validates :relevance, inclusion: { in: relevances.keys }
 
   before_validation :set_default_language, on: :create
 
   scope :open_source, lambda {
-    with_language.where(is_private: false)
+    internal.with_language.where(is_private: false)
```
santib, line 54: maybe a spec to show that open source projects don't include `ignored` repos

SandroDamilano: Actually, [this spec](https://github.com/rootstrap/rs-code-review-metrics/pull/278/files#diff-fb312d8dcf15fea2e04d19523d09ff27R74) is already doing that (although it's an easy miss, I must admit 😅 )
I create several projects in the setup with different attributes, and one of them is precisely an `ignored_project`, which I'm asserting it's not showing up as an open source project (although I'm not doing it explicitly).
If you think it's a bit obscure, I could create different contexts and explicitly assert the different type of projects are not being included

santib: I missed that one, I'm good then 👌

PR #279, opened 2020-07-31
take in count the weeks filter for project summary

hvilloria:
## What does this PR do?

* the weeks filter now affects the code climate project summary in department/project metrics views

### DIFF
--- a/app/controllers/development_metrics_controller.rb
+++ b/app/controllers/development_metrics_controller.rb
@@ -48,7 +48,7 @@ class DevelopmentMetricsController < ApplicationController
   def code_climate_department_summary
     CodeClimate::ProjectsSummaryService.call(
       department: department,
-      from: nil,
+      from: metric_params[:period],
       technologies: []
     )
   end
--- a/app/services/code_climate/projects_summary_service.rb
+++ b/app/services/code_climate/projects_summary_service.rb
@@ -56,7 +56,7 @@ module CodeClimate
     end
 
     def metrics_in_time_period
-      if from && !from.zero?
+      if from.positive?
         metrics_in_department.where(snapshot_time: from.weeks.ago..Time.zone.now)
       else
         metrics_in_department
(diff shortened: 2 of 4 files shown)


app/controllers/development_metrics_controller.rb
```diff
@@ -48,7 +48,7 @@ def code_climate_project_summary
   def code_climate_department_summary
     CodeClimate::ProjectsSummaryService.call(
       department: department,
-      from: nil,
+      from: metric_params[:period],
```
SandroDamilano, line 51: Do we really need to change the implementation of `CodeClimate::ProjectsSummaryService`? Wouldn't we just need to change parameter we're sending? If it already accepts a date, why don't we convert the input we have into a date?
We could just do something like
`from: metric_params[:period].weeks.ago,`
Or am I missing something?

app/services/code_climate/projects_summary_service.rb
```diff
@@ -56,8 +56,8 @@ def metrics_in_given_technologies
     end
 
     def metrics_in_time_period
-      if from
-        metrics_in_department.where('snapshot_time >= ?', from)
+      if from && !from.zero?
```
santib, line 59:
```suggestion
      if from.positive?
```

`from` will always be present (because fo the `.to_i`) and be zero or positive, right?

app/controllers/development_metrics_controller.rb
hvilloria: this implementation was done without counting the params that are handled in the frontend, the frontend always sent just an integer and since this sent just a `nil` it didn't crash, if you check the `CodeClimate::ProjectsDetailsService` you will notice the key `from` is an integer

app/services/code_climate/projects_summary_service.rb
hvilloria: right!

hvilloria: i will change it in project detail service too

app/controllers/development_metrics_controller.rb
SandroDamilano: Alright, I'm okay with the intention of them matching.
I should note, though, that I think now the name `from` is not very expressive, since it seems to be expecting a date instead of a number of weeks for the look-back

PR #281, opened 2020-07-31
[EM-133] Fix charts starting at the same month in Tech Blog metrics

SandroDamilano:
## What does this PR do?
It fixes the difference in the starting point shown in the Tech Blog chart, where the Blog Posts chart started a month earlier than the Visits one.

image

Resolves [133](https://rootstrap.atlassian.net/browse/EM-133)

### DIFF
--- a/app/services/builders/metric_chart/blog/technology_blog_post_count.rb
+++ b/app/services/builders/metric_chart/blog/technology_blog_post_count.rb
@@ -29,7 +29,7 @@ module Builders
         end
 
         def period_start
-          @period_start ||= periods.months.ago.beginning_of_month
+          @period_start ||= (periods - 1).months.ago.beginning_of_month
         end
       end
     end
--- a/spec/services/builders/metric_chart/blog/technology_blog_post_count_spec.rb
+++ b/spec/services/builders/metric_chart/blog/technology_blog_post_count_spec.rb
@@ -69,5 +69,18 @@ describe Builders::MetricChart::Blog::TechnologyBlogPostCount do
         expect(described_class.call.totals[:data][current_month_key]).to eq 1
       end
     end
+
+    describe 'periods returned' do
+      let(:technology) { create(:technology) }
+      let(:periods) { 13 }
+
+      before do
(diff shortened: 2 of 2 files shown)


app/services/builders/metric_chart/blog/technology_blog_post_count.rb
```diff
@@ -29,7 +29,7 @@ def new_monthly_blog_posts
         end
 
         def period_start
-          @period_start ||= periods.months.ago.beginning_of_month
+          @period_start ||= (periods - 1).months.ago.beginning_of_month
```
santib, line 32: just curious, but isn't there a more abstract way to prevent this from happening in future charts?

SandroDamilano: I'm not entirely sure what you mean. This implementation is only tied to this particular chart, so if some other one needed to do something like this we could just try to reuse the code and we would be safe.
As for the cause of this problem, it's because when I do `periods.months.ago` I'm considering `periods + 1` amount of months, while the other charts use the gem `groupdate`: `.group_by_period(grouping_period, :value_timestamp, last: periods)` which considers the current month and the `periods - 1` last ones, adding to a total of `periods`.

I could make both use the same `period_start` by changing the groupdate method options to:
`.group_by_period(grouping_period, :value_timestamp, range: period_start..Time.zone.now)`, but we would loose the simplicity of the method we have right now

santib: I meant that if we are going to be doing `x - 1` in every service we implement, maybe we could centralize that logic (maybe at the time the param is sent, nothing complex). But I don't want to make things complex, just asking because it's hard to know when `x` or `x - 1` is the correct option

SandroDamilano: Yeah, I know it's hard, and that's why I had this oversight 😅 
But this `x - 1` here is truly the exception now. This metric is the only one that falls into having to do something like this.

santib: makes sense

PR #286, opened 2020-08-04
[Feature] Highlight section that user is seeing

hvilloria:
## What does this PR do?

[Jira EM-123](https://rootstrap.atlassian.net/browse/EM-123)

usersection

### DIFF
--- a/app/views/layouts/_navigation.html.erb
+++ b/app/views/layouts/_navigation.html.erb
@@ -1,6 +1,6 @@
 <nav class="nav nav-style nav-txt-color p-2 nav-items-container col-md-12 d-flex justify-content-between">
   <div class="d-flex flex-row">
-    <%if current_page?(development_metrics_path)%>
+    <%if content_url(/development_metrics/)%>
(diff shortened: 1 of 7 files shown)


app/views/layouts/_navigation.html.erb
```diff
@@ -1,6 +1,20 @@
...
+    <%else%>
+      <%= link_to 'Tech Blog', tech_blog_path, class: 'nav-link active nav-txt-color mx-1'%>
+    <%end%>
+    <%if current_page?(open_source_path)%>
+      <%= link_to 'Open Source', open_source_path, class: 'nav-link active nav-txt-color selected mx-1'%>
+    <%else%>
+      <%= link_to 'Open Source', open_source_path, class: 'nav-link active nav-txt-color mx-1'%>
+    <%end%>
+  </div>
+  <%= link_to 'Admin', admin_events_path, class: 'nav-link active nav-txt-color mx-1'%>
```
hvilloria, line 19: im not sure if using a ternary operator would be better

santiagovidal: There's probably a DRYer way of doing it but I'm ok with this

PR #288, opened 2020-08-06
WIP: [Feature] ignore weekends for MT/RT

hvilloria:
## What does this PR do?

* substract every saturday/sunday as seconds when calculating mt/rt distributions


app/services/builders/review_turnaround.rb
```diff
@@ -11,7 +13,17 @@ def call
...
 
     def calculate_turnaround
-      @review_request.reviews.first.opened_at.to_i - @review_request.created_at.to_i
+      (opened_review_request.to_i - @review_request.created_at.to_i) - weekend_days_as_seconds
+    end
+
+    def weekend_days_as_seconds
+      range_days = @review_request.created_at.to_date..@review_request.reviews.first.opened_at.to_date
+      weekends = range_days.select { |day| day.wday == 6 || day.wday ==0 }.count
+      weekends * SECONDS_IN_A_DAY
```
santib, line 22: not sure if we care about it but if someone submits a review on a saturday or sunday this might not work fine

app/services/builders/review_turnaround.rb
```diff
@@ -11,7 +13,17 @@ def call
     private
 
     def calculate_turnaround
-      @review_request.reviews.first.opened_at.to_i - @review_request.created_at.to_i
+      (opened_review_request.to_i - @review_request.created_at.to_i) - weekend_days_as_seconds
+    end
+
+    def weekend_days_as_seconds
+      range_days = @review_request.created_at.to_date..@review_request.reviews.first.opened_at.to_date
```
SandroDamilano, line 20: We could use `opened_review_request` here (and probably memoize it)

spec/services/builders/review_turnaround_spec.rb
```diff
@@ -7,26 +7,58 @@
     let(:santib) { create(:user, login: 'santib') }
     let(:pr) { create(:pull_request, owner: vita, project: project) }
 
-    let(:rr) do
-      create(:review_request, owner: vita, reviewer: santib, project: project, pull_request: pr)
+    before do
+      allow_any_instance_of(Events::Review).to receive(:build_review_turnaround)
```
SandroDamilano, line 11: Wouldn't `Events::Review.skip_callback(:create, :after, :build_review_turnaround)` work? It seems more _intention revealing_

hvilloria: the `skip_callback` is dangerous 😆 , when i used it i didn't know that method works across the tests i mean, it will skip all the callbacks for everytime that a event creation is invoked, that's why i changed by a mock

SandroDamilano: Oh, I didn't know that, good call!

SandroDamilano: If you asked me, these are the things why I don't like these hooks, and I'd rather do without them.
But no, you didn't ask me 😂

hvilloria: i totally agree with you,  sometimes it is really confusing 😆

app/services/builders/review_turnaround.rb
```diff
@@ -11,7 +13,17 @@ def call
     private
 
     def calculate_turnaround
-      @review_request.reviews.first.opened_at.to_i - @review_request.created_at.to_i
+      (opened_review_request.to_i - @review_request.created_at.to_i) - weekend_days_as_seconds
+    end
+
+    def weekend_days_as_seconds
```
santiagovidal, line 19: We also want this for the Merge Time calculations, so perhaps this logic should go somewhere else, where it can be reused.

santiagovidal:
A non-blocking suggestion: I think it would be better to have a method that returns the number of seconds between two timestamps, excluding weekends. With your current implementation, it feels like the logic of excluding weekends is split between the builder class and `DistributionInterval`. One does not make sense without the other.

hvilloria:
@santib @santiagovidal i changed it a little, now a `weekend_seconds_interval` class is used to obtain the complete seconds to be substracted in the MT/RT calculations, i dont know if this has more readability than previous approach and if this something more similar to your comments, let me know what do you think 🙏

app/services/weekend_seconds_interval.rb
```diff
@@ -0,0 +1,24 @@
+class WeekendSecondsInterval < BaseService
+  SECONDS_IN_A_DAY = 864_00
+
+  def initialize(start_date:, end_date:)
+    @start_date = start_date
+    @end_date = end_date
+  end
+
+  def call
+    weekends_as_seconds = weekends_count * SECONDS_IN_A_DAY
```
santib, line 10: I think this service doesn't support partial dates. e.g. if I call it like:
```
irb(main):011:0> a
=> Sun, 09 Aug 2020 19:18:43 UTC +00:00
irb(main):012:0> b = 5.hours.ago(a)
=> Sun, 09 Aug 2020 14:18:43 UTC +00:00
irb(main):013:0> WeekendSecondsInterval.call(start_date: a, end_date: Time.current)
=> 86400
irb(main):014:0> WeekendSecondsInterval.call(start_date: b, end_date: Time.current)
=> 86400
```

If I'd have had to count the amount of seconds between two dates without weekends, I'd have divided it in 3 parts:
1. seconds from start time til end of first day (unless it's a weekend)
2. seconds from beginning of last day til end time (unless it's a weekend)
3. seconds of full days between start time and end time without weekends

you can get the intermediate days like this `(start_date.tomorrow.to_date..end_date.yesterday.to_date).select { |day| day.on_weekday? }.size * SECONDS_IN_A_DAY`

and then `start_date.on_weekday? start_date.end_of_day - start_date : 0` and `end_date.on_weekday? end_date - end_date.beginning_of_day : 0`

and the sum of the 3 parts is the result.

but this solution might be fine for now

app/services/weekend_seconds_interval.rb
```diff
@@ -0,0 +1,24 @@
+class WeekendSecondsInterval < BaseService
+  SECONDS_IN_A_DAY = 864_00
```
santib, line 2: shouldn't this be `86_400` ?

app/services/weekend_seconds_interval.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def call
+    weekends_as_seconds = weekends_count * SECONDS_IN_A_DAY
+    return weekends_as_seconds unless @end_date.saturday? || @end_date.sunday?
+
+    weekends_as_seconds - seconds_of_last_weekend_day
+  end
+
+  def seconds_of_last_weekend_day
+    @end_date.end_of_day.to_i - @end_date.to_i
```
santib, line 17: are `.to_i` calls necessary?

hvilloria: it would be `.to_i` or a `round`

app/services/weekend_seconds_interval.rb
hvilloria: i will refactor it then 👍

PR #290, opened 2020-08-07
Make open source jobs catch api errors

SandroDamilano:
## What does this PR do?
It prevents the job `OpenSourceMetricsUpdaterJob` from ending in failure due to any single failed API call made to Github.
If such an error may occur, it will only affect the problematic project and the rest will be able to be updated successfully. The error occurrence will also be tracked by Exception Hunter.

### DIFF
--- a/app/services/github_repository_client.rb
+++ b/app/services/github_repository_client.rb
@@ -26,7 +26,9 @@ class GithubRepositoryClient
         request.headers['Accept'] = 'application/vnd.github.v3.raw'
       end
 
-      return response.body if response.success?
+      return response.body
+    rescue Faraday::ResourceNotFound
+      next
     end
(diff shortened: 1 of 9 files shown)


app/services/github_repository_client.rb
```diff
@@ -26,14 +26,17 @@ def find_in_locations
...
+    rescue Faraday::ResourceNotFound
+      next
     end
     ''
   end
 
   def connection
     Faraday.new('https://api.github.com') do |conn|
       conn.basic_auth(ENV['GITHUB_ADMIN_USER'], ENV['GITHUB_ADMIN_TOKEN'])
+      conn.response :raise_error
```
martinjaimem, line 39: 👏

PR #291, opened 2020-08-07
Make blog jobs catch api errors

SandroDamilano:
## What does this PR do?
It prevents all blog-related jobs from ending in failure due to any single failed API call made to the Wordpress API*.
If such an error may occur, it will only affect the problematic blog post and the rest will be able to be updated successfully. The error occurrence will also be tracked by Exception Hunter.

*The only exception being a failed request to the `me/posts` endpoint, which retrieves all blog posts. If it failed, no update could be made anyway, so we'll let the job fail

### DIFF
--- a/app/services/processors/blog_post_views_updater.rb
+++ b/app/services/processors/blog_post_views_updater.rb
@@ -14,6 +14,8 @@ module Processors
           create_metric(year, month, views)
         end
       end
+    rescue Faraday::Error => exception
+      ExceptionHunter.track(exception)
     end
 
     private
--- a/app/services/processors/blog_posts_updater.rb
+++ b/app/services/processors/blog_posts_updater.rb
@@ -2,13 +2,12 @@ module Processors
   class BlogPostsUpdater < BaseService
     def call
       blog_posts.each do |blog_post_payload|
-        blog_post_id = blog_post_payload[:ID]
-        blog_post = BlogPost.find_or_initialize_by(blog_id: blog_post_id)
+        blog_post = BlogPost.find_or_initialize_by(blog_id: blog_post_payload[:ID])
         blog_post.published_at = blog_post_payload[:date]
         blog_post.slug = blog_post_payload[:slug]
         blog_post.status = blog_post_payload[:status]
         blog_post.url = blog_post_payload[:URL]
-        blog_post.technologies = technologies_for(blog_post_id)
+        blog_post.technologies = technologies_for(blog_post)
 
         blog_post.save!
       end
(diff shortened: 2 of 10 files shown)


app/services/wordpress_service.rb
```diff
@@ -55,15 +47,18 @@ def request_access_token
...
+    response = Faraday.post('https://public-api.wordpress.com/oauth2/token', request_body)
     response_body = JSON.parse(response.body)
     raise_invalid_token_error(response_body) if response.status == 400
 
     response_body['access_token']
   end
 
   def connection
-    Faraday.new('https://public-api.wordpress.com/')
+    Faraday.new('https://public-api.wordpress.com/rest/v1.1/') do |conn|
```
martinjaimem, line 58: maybe we should move the API url to a constant or env variable, WDYT?

SandroDamilano: yeah, absolutely!

app/services/processors/blog_posts_updater.rb
```diff
@@ -24,9 +23,12 @@ def categorizer
...
     end
 
-    def technologies_for(blog_post_id)
-      blog_post_payload = wordpress_service.blog_post(blog_post_id)
+    def technologies_for(blog_post)
+      blog_post_payload = wordpress_service.blog_post(blog_post.blog_id)
       categorizer.technologies_for(blog_post_payload)
+    rescue Faraday::Error => exception
+      ExceptionHunter.track(exception)
+      blog_post.technologies
```
santiagovidal, line 31: I'm not sure what this line is for.

SandroDamilano: It's because we care about the return value of this method. It should return the technologies that match a blog post, so that we can assign them to it (check [here](https://github.com/rootstrap/rs-code-review-metrics/pull/291/files#diff-a9bfba1ad10a239e09f1b608d16cd788R10))
If we couldn't get the information from the API to know which ones we should assign, at least we will be assigning the ones the blog post already had, allowing them to remain untouched.
So if the blog post had already technologies associated, they'll remain the same.
And if the blog post is new, we'll leave it without any technology (we could assign the `other` technology in this case, though).

santiagovidal: Makes total sense 👍 my bad

app/services/processors/blog_post_views_updater.rb
```diff
@@ -14,6 +14,8 @@ def call
           create_metric(year, month, views)
         end
       end
+    rescue Faraday::Error => exception
```
santib, line 17: doesn't ExceptionHunter track exceptions automatically? 🤔 cc @brunvez

brunvez: yeah it should automatically track exceptions on jobs and if they are not handled on every request

SandroDamilano: It certainly does! But it doesn't if we catch the error first 😅

PR #294, opened 2020-08-11
[EM-135] Import rootstrap projects

SandroDamilano:
## What does this PR do?
It creates a job to import all Rootstrap projects from Github. It will add the new ones and update the old ones.
The job is scheduled to run every Saturday morning

Resolves [#135](https://rootstrap.atlassian.net/browse/EM-135)


app/services/code_owners/request.rb
```diff
@@ -2,7 +2,7 @@ module CodeOwners
   class Request < BaseService
     def call
       Project.find_each do |project|
-        content_file = GithubRepositoryClient.new(project).code_owners
+        content_file = GithubClient::Repository.new(project).code_owners
```
hdamico, line 5: 👍

app/services/github_client/organization.rb
```diff
@@ -0,0 +1,21 @@
+module GithubClient
+  class Organization < GithubClient::Base
+    MAX_REPOSITORIES_PER_PAGE = 100
+
+    def repositories(acc_repositories = [], page_number = 1)
+      request_params = {
+        'page': page_number,
```
santib, line 7: `page: page_number` doesn't work?

SandroDamilano: You're right! I'll make the change now

PR #297, opened 2020-08-12
[EM-139.2] Deprecate blog post count metrics

SandroDamilano:
## What does this PR do?
It deprecates all code and data related to the `blog_post_count` metrics, as now we won't use the `metrics` table for storing such data anymore.

To delete the data, a task `metrics:remove_blog_post_count` has been created. It deletes all `blog_post_count` rows from the `metrics` table

Resolves [#139](https://rootstrap.atlassian.net/browse/EM-139)


app/services/processors/blog_post_count_metrics_remover.rb
```diff
@@ -0,0 +1,9 @@
+module Processors
+  class BlogPostCountMetricsRemover < Processors::Base
+    private
+
+    def process
+      Metric.where(name: Metric.names[:blog_post_count]).destroy_all
+    end
+  end
+end
```
hdamico, line 9: Is this something that are we going to use more than once? Otherwise, why don't you put the code inside the rake?

spec/lib/metrics_remove_blog_post_count_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+require 'rake'
```
horacio, line 2: Is this actually needed?

app/services/processors/blog_post_count_metrics_remover.rb
SandroDamilano: I hadn't thought about that difference, I just always assumed it was better to just write the code outside of the task.
Indeed it's gonna be run just once. I planned to remove this code and the task as well anyway when I finished using it, though. Should I still move the code?

spec/lib/metrics_remove_blog_post_count_spec.rb
SandroDamilano: It is, because of the rake utils I'm using in the spec

PR #305, opened 2020-08-19
Refactor error handling for CodeClimateMetricsUpdaterJob

SandroDamilano:
## What does this PR do?
It refactors error handling concerning Code Climate API requests.
We used to just ignore all API errors, which resulted in a lot of `nil` checks everywhere.
The idea now is to let Faraday fail and raise the correspondent exception, so we can forget about all those `nil`s and handle the exceptions only where we need to.


hvilloria:
👏

PR #306, opened 2020-08-19
[EM-151] Change code climate repository endpoint

SandroDamilano:
## What does this PR do?
It adds a new way to get the project info from Code Climate.
We used to get it using the project name, but if the name changed on Github and not in Code Climate, we wouldn't be able to retrieve the info anymore.
[As the docs for Code Climate state](https://developer.codeclimate.com/#get-repository), there are two ways to get this info: using the project name or using the repository id. So, to avoid the problem described above, we will opt to use the endpoint with the repository id, which we know can't change.

To do so, we now:
- Start saving this repository id in our own database (by the name of `cc_repository_id` in `CodeClimateProjectMetric`)
- Use it (when available) to get the project info from Code Climate

**Note**: The only way to get this repository id is by hitting the Code Climate repository endpoint with the project name. So the first time we query any project info from Code Climate we will be forced to do so with that endpoint anyway.

Resolves [#151](https://rootstrap.atlassian.net/browse/EM-151)

### DIFF
--- a/app/services/code_climate/get_project_summary.rb
+++ b/app/services/code_climate/get_project_summary.rb
@@ -1,15 +1,31 @@
 module CodeClimate
   class GetProjectSummary < BaseService
-    def initialize(github_slug:)
-      @github_slug = github_slug
+    CODE_CLIMATE_API_ORG_NAME = 'rootstrap'.freeze
+
+    def initialize(project:)
+      @project = project
     end
 
     def call
       repository&.summary
     end
 
+    private
+
+    attr_reader :project
+
     def repository
-      CodeClimate::Api::Client.new.repository_by_slug(github_slug: @github_slug)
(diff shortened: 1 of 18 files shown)


app/services/code_climate/get_project_summary.rb
```diff
@@ -1,15 +1,31 @@
 module CodeClimate
   class GetProjectSummary < BaseService
-    def initialize(github_slug:)
-      @github_slug = github_slug
+    CODE_CLIMATE_API_ORG_NAME = 'rootstrap'.freeze
```
hdamico, line 3: Shouldn't this be an environment variable?

SandroDamilano: You're totally right, it should!
As it was already like that before this PR, would you be fine if I changed it in another PR? I can probably start working on it right away

PR #307, opened 2020-08-19
Fix duplicate blog posts flaky test

SandroDamilano:
## What does this PR do?
It just prevents a quite improbable flaky test to trigger.
(Note that *quite improbable* is by no means *impossible*, as it did trigger for me in a test run 😂 )

### DIFF
--- a/spec/factories/blog_posts.rb
+++ b/spec/factories/blog_posts.rb
@@ -22,7 +22,7 @@
 #
 FactoryBot.define do
   factory :blog_post do
-    blog_id { Faker::Number.number(digits: 4) }
+    sequence(:blog_id)
     slug { 'ruby-is-awesome' }
     url { 'https://www.rotstrap.com/blog/ruby_is_awesome' }
     status { BlogPost.statuses[:publish] }


spec/factories/blog_posts.rb
```diff
@@ -22,7 +22,7 @@
 #
 FactoryBot.define do
   factory :blog_post do
-    blog_id { Faker::Number.number(digits: 4) }
+    blog_id { Faker::Number.unique.number(digits: 4) }
```
santib, line 25: when I need unique things I usually prefer going with `SecureRandom` module because the problem with Faker is that it can run out of trials and you get `Faker::UniqueGenerator::RetryLimitExceeded`

SandroDamilano: Yeah, I'm aware of that issue with `Faker`. I didn't know I could use `SecureRandom` for this purpose, how should I use it?

santib: well now that I checked https://ruby-doc.org/stdlib-2.5.1/libdoc/securerandom/rdoc/SecureRandom.html I realized that there is no method to get just a random number with x amount of digits 😅 

other option would be using FactoryBot's `sequence` https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences . But it's up to you changing it or not

SandroDamilano: Oh, right! I remember looking into that option before. I think it's the best option here. Thank you!

PR #311, opened 2020-08-24
[EM-137] Update projects when importing them

SandroDamilano:
## What does this PR do?
New projects were being correctly imported from Github, but we were not updating those that were previously in our DB.
This PR fixes that.

Resolves [#137](https://rootstrap.atlassian.net/browse/EM-137)

### DIFF
--- a/app/services/builders/project.rb
+++ b/app/services/builders/project.rb
@@ -11,12 +11,14 @@ module Builders
     private
 
     def fetch_or_create
-      ::Project.find_or_create_by!(github_id: @repository_data['id']) do |project|
-        project.name = @repository_data['name']
-        project.description = @repository_data['description']
-        project.is_private = @repository_data['private']
-        project.relevance = ::Project.relevances[:ignored] if @repository_data['archived']
-      end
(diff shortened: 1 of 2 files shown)


app/services/builders/project.rb
```diff
@@ -11,12 +11,14 @@ def call
...
-        project.relevance = ::Project.relevances[:ignored] if @repository_data['archived']
-      end
+      project = ::Project.find_or_initialize_by(github_id: @repository_data['id'])
+      project.name = @repository_data['name']
+      project.description = @repository_data['description']
+      project.is_private = @repository_data['private']
+      project.relevance = ::Project.relevances[:ignored] if @repository_data['archived']
+      project.save!
+
+      project
```
hdamico, line 21: I believe we can use `.tap`.
`::Project.find_or_initialize_by(github_id: @repository_data['id']).tap |project|`
It not relevant anyway.

SandroDamilano: Absolutely! Nice catch!

PR #313, opened 2020-08-26
[EM-143] Show overview info for department projects

SandroDamilano:
## What does this PR do?
It adds an overview to the department metrics page showing the amount of projects per language and relevance within that department.

image (2)

Resolves [#143](https://rootstrap.atlassian.net/browse/EM-143)

### DIFF
--- a/app/services/builders/department_overview.rb
+++ b/app/services/builders/department_overview.rb
@@ -7,11 +7,9 @@ module Builders
     end
 
     def call
-      department_overview = language_names.each_with_object({}) do |language, overview|
+      language_names.each_with_object({}) { |language, overview|
         overview[language] = language_overview(language)
-      end
-
-      department_overview.with_indifferent_access
+      }.with_indifferent_access
     end
 
     private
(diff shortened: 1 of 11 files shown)


app/services/builders/department_overview.rb
```diff
@@ -0,0 +1,57 @@
...
+    def initialize(department)
+      @department = department
+    end
+
+    def call
+      department_overview = language_names.each_with_object({}) do |language, overview|
+        overview[language] = language_overview(language)
+      end
+
+      department_overview.with_indifferent_access
```
horacio, line 14: Just sharing another shorter option...

```suggestion
      department_overview = language_names.index_with { |language|
        language_overview(language)
      }.with_indifferent_access
```

app/services/builders/department_overview.rb
```diff
@@ -0,0 +1,57 @@
...
+      {
+        per_relevance: relevances_overview,
+        totals: relevances_overview.values.sum
+      }
+    end
+
+    def relevances_overview(language)
+      project_relevances.each_with_object({}) do |relevance, overview|
+        overview[relevance] = projects_by_language_and_relevance[[language, relevance]] || 0
+      end
```
horacio, line 42: Here I think we can do the same as above using `index_with`...

```suggestion
      project_relevances.index_with do |relevance|
        projects_by_language_and_relevance[[language, relevance]] || 0
      end
```

app/services/builders/department_overview.rb
SandroDamilano: True! I like it!

app/services/builders/department_overview.rb
SandroDamilano: Oh, I didn't know about that method! Thank you!

PR #318, opened 2020-09-02
[EM-146.1] Add PR Size metric

SandroDamilano:
## What does this PR do?
It creates the `PullRequestSize` model, which will contain the information about the PR Size metric.
It will be generated every time a new `PullRequest` is created.

Resolves [#146](https://rootstrap.atlassian.net/browse/EM-146)

### DIFF
--- a/app/models/events/pull_request.rb
+++ b/app/models/events/pull_request.rb
@@ -46,6 +46,7 @@ module Events
                        inverse_of: :pull_request
     has_many :events, as: :handleable, dependent: :destroy
     has_one :merge_time, dependent: :destroy
+    has_one :pull_request_size, dependent: :destroy
 
     validates :state, inclusion: { in: states.keys }
     validates :github_id,
--- a/app/services/github_client/base.rb
+++ b/app/services/github_client/base.rb
@@ -1,11 +1,29 @@
 module GithubClient
   class Base
+    private
+
     def connection
       Faraday.new('https://api.github.com') do |conn|
         conn.basic_auth(ENV['GITHUB_ADMIN_USER'], ENV['GITHUB_ADMIN_TOKEN'])
         conn.headers['Accept'] = 'application/vnd.github.v3+json'
         conn.response :raise_error
       end
+    end
+
+    def get_all_paginated_items(url, max_per_page, accumulated_items = [], page_number = 1)
+      request_params = {
+        page: page_number,
+        per_page: max_per_page
+      }
+
+      response = connection.get(url) do |request|
+        request.params = request_params
+      end
+      new_items = JSON.parse(response.body).map(&:with_indifferent_access)
+
+      return accumulated_items if new_items.empty?
+
+      get_all_paginated_items(url, max_per_page, accumulated_items + new_items, page_number + 1)
     end
   end
 end
--- a/app/services/github_client/base.rb
+++ b/app/services/github_client/base.rb
@@ -10,7 +10,7 @@ module GithubClient
       end
     end
 
-    def get_all_paginated_items(url, max_per_page, accumulated_items = [], page_number = 1)
+    def get_all_paginated_items(url, max_per_page, accumulated_items: [], page_number: 1)
       request_params = {
         page: page_number,
         per_page: max_per_page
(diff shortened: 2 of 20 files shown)


app/models/events/pull_request.rb
```diff
@@ -46,6 +46,7 @@ class PullRequest < ApplicationRecord
                        inverse_of: :pull_request
     has_many :events, as: :handleable, dependent: :destroy
     has_one :merge_time, dependent: :destroy
+    has_one :pull_request_size, dependent: :destroy
```
santib, line 49: why isn't this an integer column?

app/services/github_client/base.rb
```diff
@@ -1,11 +1,29 @@
...
+
     def connection
       Faraday.new('https://api.github.com') do |conn|
         conn.basic_auth(ENV['GITHUB_ADMIN_USER'], ENV['GITHUB_ADMIN_TOKEN'])
         conn.headers['Accept'] = 'application/vnd.github.v3+json'
         conn.response :raise_error
       end
     end
+
+    def get_all_paginated_items(url, max_per_page, accumulated_items = [], page_number = 1)
```
santib, line 13: minor comment but maybe `max_per_page, accumulated_items = [], page_number = 1` should be keyword arguments because they look more like options than to ordinary arguments to me.. thoughts? also max_per_page could have a default value too, right?

db/migrate/20200901185355_create_pull_request_sizes.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePullRequestSizes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :pull_request_sizes do |t|
+      t.references :pull_request, null: false, foreign_key: true
+      t.integer :value
```
santib, line 5: maybe `null: false` ?

santib (approved):
👏

app/models/events/pull_request.rb
SandroDamilano: Well, that's a great question.
When I was building this, I noticed that the other two distribution metrics had their own separated model (`MergeTime` and `CompletedReviewTurnaround`). So I made the decision of creating one for this new metric just to stick with the pattern and make it more clear, even when it's less efficient (in terms of performance and storage).
I'm not too sure about that decision, though, and it's easy to try the other option, so I'd love to hear your thoughts!

santib: I see.. I like to idea of keeping things consistent. I guess there is a reason why we have models instead of just values 🤔 do you remember why was it @santiagovidal @hvilloria ?

app/services/github_client/base.rb
SandroDamilano: I think you may be right about the keyword arguments.
As for the default value, it's not that straightforward. Each API could have it's own max value, so placing a default value would assume they are all the same ones, which [Github specifically says it may not be the case](https://docs.github.com/en/rest/guides/traversing-with-pagination). So I'd rather leave that one as it is.

db/migrate/20200901185355_create_pull_request_sizes.rb
SandroDamilano: Yeah, it makes sense!

app/models/events/pull_request.rb
SandroDamilano: I don't quite remember if we considered it at that moment, but I guess they weren't as straightforward as this one.
Both review turnaround and merge time don't occur at the same time as the PR creation for obvious reasons, so we would need to store when the PR was merged or reviewed as well, or to consider the time the PR was opened for it.
Still, I can't think of a strong reason against doing it that way 🤔 
I think @hdamico came up with the idea in the first place, so maybe he could also share his input?

hdamico: Having MergeTime and ReviewTurnaround in different tables is because these two metrics are calculated between two records, so I don't think those have to go in a table column also, when it comes to query them, it is easier to get the data that we want. I don' t really remember if this is a pattern, I only remember doing this for convenience. Maybe @santiagovidal or @hvilloria remember something

hdamico: If you guys believe having ReviewTurnaround and MergeTime in a column inside PR table im good with that.

santib: But the MergeTime belongs to a PullRequest, so it could be a value in that table. And the ReviewTurnaround belongs to a ReviewRequest so the value could be just there.

That said, I'm ok keeping things this way for now. If at some point we see a benefit to do the change we can discuss it further

PR #319, opened 2020-09-04
[EM-146.2 & EM-162] Show PR Size metric

SandroDamilano:
## What does this PR do?
It adds the distribution chart for the new PR Size metric.
It also adds titles for the metrics and separators.
  
image

Resolves [#146](https://rootstrap.atlassian.net/browse/EM-146) and [#162](https://rootstrap.atlassian.net/browse/EM-162)


app/services/metrics/interval_resolver.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    def call
+      matching_interval || upper_interval
+    end
+
+    private
+
+    def matching_interval
+      ranges.find { |_interval, hour| @value < hour }&.first&.to_s
```
santiagovidal, line 14: `hour` should be renamed to something more generic, right? It can refer to time or PR size.

santiagovidal (approved):
Nice 👏 Small comment.

app/services/metrics/interval_resolver.rb
SandroDamilano: Absolutely! Thank you for noticing it!

app/services/builders/chartkick/department_distribution_data.rb
```diff
@@ -10,21 +10,52 @@ def call
...
-        merge_times
+      def resolve_interval(entity)
+        metric.resolve_interval(entity)
       end
 
-      def review_turnarounds
-        ::CompletedReviewTurnaround.joins(review_request: { project: { language: :department } })
-                                   .where(departments: { id: @entity_id })
-                                   .where(created_at: @query[:value_timestamp])
+      class MergeTime
```
santib, line 31: why do you prefer to have 3 classes in a single file instead of 1 class per file?

app/views/development_metrics/departments.erb
```diff
@@ -26,32 +26,36 @@
...
-        <% if @review_turnaround && @merge_time %>
-          <div class="row">
-            <div class="metrics">
-              <div class="graph">
-                <%= render 'development_metrics/review_turnaround/main_metric', review_turnaround: @review_turnaround %>
-              </div>
-              <div class="graph details">
-                <%= render 'development_metrics/review_turnaround/details_metric', review_turnaround: @review_turnaround %>
-              </div>
+        <% if @review_turnaround && @merge_time && @pull_request_size %>
```
santib, line 29: when are any of these 3 going to not be present? is it possible that only one of them is not preset but the other two are? in that case should we show something instead?

spec/services/metrics/interval_resolver_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    context 'when the value is bigger than all the intervals' do
+      let(:value) { 200 }
+
+      it 'returns the upper limit of the ranges with a "+" beside it' do
+        expect(interval).to eq('200+')
+      end
+    end
+  end
+
+  describe Metrics::IntervalResolver::Time do
```
santib, line 30: why do you like having 1 test file for two different classes?

app/services/builders/chartkick/department_distribution_data.rb
santib: https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#project-structure
https://github.com/fxn/zeitwerk#file-structure

SandroDamilano: I did it just for readability, but you're right, I should move them to their own files.

app/views/development_metrics/departments.erb
SandroDamilano: It's a good question. I'll check it and change it if it does not make sense.

spec/services/metrics/interval_resolver_spec.rb
SandroDamilano: The same as the other answer, but I'll just move it.

app/views/development_metrics/departments.erb
SandroDamilano: As how things work right now, these three can be either all empty or all not empty, simultaneously. And they are empty when no department has been selected.
I think we can safely change it to check only if the department has been selected.
Good catch! 👏

santib (approved):

PR #323, opened 2020-09-09
[EM-165] Update pull request size on every push

SandroDamilano:
## What does this PR do?
It catches all `push` events and creates a `Push` record for each one, associating it to its project, user and pull request (if any). If there's a pull request, its pull request size metric is updated 

Resolves [#165](https://rootstrap.atlassian.net/browse/EM-165)

### DIFF
--- a/app/services/action_handlers/push.rb
+++ b/app/services/action_handlers/push.rb
@@ -1,0 +1,17 @@
+module ActionHandlers
+  class Push < ActionHandler
+    private
+
+    def handleable?
+      true
+    end
+
+    def created
+      pull_request = @entity.pull_request
+
+      return unless pull_request
+
+      Builders::PullRequestSize.call(pull_request)
+    end
+  end
+end
--- a/db/migrate/20200908121903_add_branch_to_pull_requests.rb
+++ b/db/migrate/20200908121903_add_branch_to_pull_requests.rb
@@ -1,0 +1,5 @@
+class AddBranchToPullRequests < ActiveRecord::Migration[6.0]
+  def change
+    add_column :pull_requests, :branch, :string
+  end
+end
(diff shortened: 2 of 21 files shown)


db/migrate/20200908121903_add_branch_to_pull_requests.rb
```diff
@@ -0,0 +1,5 @@
+class AddBranchToPullRequests < ActiveRecord::Migration[6.0]
+  def change
+    add_column :pull_requests, :branch, :text
```
horacio, line 3: Just curious, why `text` instead of `string`?

SandroDamilano: Nothing in particular. Would you recommend using string?

horacio: I think they're effectively pretty similar, so I'll use whatever it's being used for this kind of values ✨

SandroDamilano: They are both being used... although `string` seems to be more popular :P
I'll see if I can quickly change it

horacio (approved):

app/services/action_handlers/push.rb
```diff
@@ -0,0 +1,17 @@
+module ActionHandlers
+  class Push < ActionHandler
+    private
+
+    def handleable?
```
hdamico, line 5: What is this for?

hdamico (approved):
Super clear!!

app/services/action_handlers/push.rb
SandroDamilano: The superclass, `ActionHandler` has the `call` method defined like this:
```
def call
    return unless handleable?

    handle_action
end
```
The rest of the subclasses use it to check if the action is handleable, but for push events there's only one possible event, which is already being handled. So for them, this method should always return `true`

PR #325, opened 2020-09-10
[EM-167] Backfill PR Size metrics

SandroDamilano:
## What does this PR do?
It adds a task to backfill the PR Size metrics for all PRs 

Task: `rake metrics:update_pr_size`

Resolves [#167](https://rootstrap.atlassian.net/browse/EM-167)


hvilloria (approved):

app/services/processors/pull_request_size_updater.rb
```diff
@@ -0,0 +1,9 @@
+module Processors
+  class PullRequestSizeUpdater < BaseService
+    def call
+      Events::PullRequest.find_each(batch_size: 100).lazy.each do |pull_request|
```
horacio, line 4: Interesting! Just out of curiosity: Why have we chosen such a small batch size? What leverage does `lazy` provides here?

spec/lib/metrics_update_pr_size_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:rake_update_pr_size) do
+    Rake.application.invoke_task 'metrics:update_pr_size'
+  end
+
+  it 'runs the PullRequestSizeUpdaterJob' do
+    expect(PullRequestSizeUpdaterJob).to receive(:perform_now)
+
+    rake_update_pr_size
+  end
+end
```
horacio, line 23: Love this

horacio (approved):
Looking great!

app/services/processors/pull_request_size_updater.rb
SandroDamilano: Good point, that batch size is indeed too low. I'd better use the default size of 1000 instead.
As for `lazy`... it's a good question. I remember @hdamico suggesting me to add it in another query some time ago, so I thought it would be suitable here too. But yeah, I'm not entirely sure what difference it makes.

PR #328, opened 2020-09-14
[EM-178] Stop using created_at for time reference in merge time and PR size

SandroDamilano:
## What does this PR do?
`MergeTime` and `PullRequestSize` metrics use their own `created_at` attribute to reference the time those metrics apply to. But that attribute may not always reflect that time accurately, and it's not its primary purpose to do it either.
This PR makes them use better references:
- `MergeTime` uses now `merged_at` in `PullRequest`
- `PullRequestSize` uses now `opened_at` in `PullRequest`

Resolves [#178](https://rootstrap.atlassian.net/browse/EM-178)

### DIFF
--- a/app/services/builders/chartkick/department_distribution_data.rb
+++ b/app/services/builders/chartkick/department_distribution_data.rb
@@ -10,10 +10,7 @@ module Builders
       private
 
       def retrieve_records
-        metric
-          .records_with_departments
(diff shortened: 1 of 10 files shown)


hvilloria (approved):

santib (approved):
👏

PR #329, opened 2020-09-14
Metrics for merge time 

hvilloria:
## What does this PR do?
It uses MergeTime model as parameter to calculate metrics for every entity. This fixes the merge time average metrics bug. 

## Hosward notes
* this branch was initially created to resolve the issue for week metrics values but i could not finish it
* the main idea was to refactor the calculation because i thought the independency between de distribution and the metric calculation was the problem but it was not.
* after the refactor all the information i collected is in this [Jira](https://rootstrap.atlassian.net/browse/EM-155), i hope it be helpful

### DIFF
--- a/app/services/metrics/merge_time/per_user_project.rb
+++ b/app/services/metrics/merge_time/per_user_project.rb
@@ -21,7 +21,7 @@ module Metrics
 
       def filtered_merge_times
         ::MergeTime.joins(:pull_request)
-                   .where(created_at: merge_time_interval)
+                   .where(pull_requests: { merged_at: merge_time_interval })
                    .group(:project_id, :owner_id)
                    .pluck(:project_id, :owner_id, Arel.sql('AVG(merge_times.value)'))
       end
--- a/app/services/metrics/merge_time/per_user_project.rb
+++ b/app/services/metrics/merge_time/per_user_project.rb
@@ -5,17 +5,11 @@ module Metrics
 
       def process
         ActiveRecord::Base.transaction do
-          merge_times_in_batches do |project_id, owner_id, merge_time_value|
+          filtered_merge_times.each do |project_id, owner_id, merge_time_value|
             entity = find_user_project(owner_id, project_id)
             create_or_update_metric(entity.id, UsersProject.name,
                                     merge_time_interval, merge_time_value, :merge_time)
           end
-        end
-      end
-
-      def merge_times_in_batches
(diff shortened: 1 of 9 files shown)


spec/services/metrics/merge_time/per_department_spec.rb
```diff
@@ -2,44 +2,93 @@
...
       before do
-        create(:metric, value: 30.minutes, ownable: first_project, name: :merge_time)
-        create(:metric, value: 25.minutes, ownable: second_project, name: :merge_time)
+        first_project_pull_request.update!(merged_at: beginning_of_day + 1.hour)
+        second_project_pull_request.update!(merged_at: beginning_of_day + 2.hours)
       end
 
       it 'creates two metrics' do
-        expect { described_class.call }.to change { Metric.count }.from(2).to(4)
+        expect { described_class.call }.to change { Metric.count }.from(0).to(2)
```
santib, line 34:
```suggestion
        expect { described_class.call }.to change { Metric.count }.by(2)
```
?

santib (approved):
Great job 💯

app/services/metrics/merge_time/per_user_project.rb
```diff
@@ -1,42 +1,38 @@
...
           end
-          calculate_metrics_avg(entities, :merge_time)
         end
       end
 
-      def pull_requests_in_batches
-        filtered_pull_requests.find_each(batch_size: BATCH_SIZE).lazy.each do |pull_request|
-          yield(pull_request)
+      def merge_times_in_batches
+        filtered_merge_times.each do |project_id, owner_id, merge_time_value|
```
hdamico, line 17: Why did you stop using BATCH_SIZE and `lazy`?

hdamico (approved):

app/services/metrics/merge_time/per_user_project.rb
SandroDamilano: I don't have an answer for that haha
@hvilloria do you? Otherwise I'll just put them back

SandroDamilano: Sorry, I do have an answer.
`filtered_merge_times` no longer returns an `ActiveRecord_Relation`, it returns an `Array`. So we can't use `find_each` anymore.
Given this, I'll just delete this `merge_times_in_batches` method, as it's useless now

PR #332, opened 2020-09-22
Fix/em 169 validation error

mrubi-rootstrap:
## What does this PR do?

It addresses [EM-169 - ActiveRecord::RecordNotUnique in pull_request_size](https://rootstrap.atlassian.net/browse/EM-169)

## Description

[ExceptionHunter](http://engineering-metrics.herokuapp.com/exception_hunter/errors/4) reported the error `ActiveRecord::RecordNotUnique` during the execution of PullRequestSize.call

The source of the error may be 2 concurrent processes making an update on the same pull request at the same time as shown in
```ruby
::PullRequestSize.find_or_initialize_by(pull_request: pull_request).tap do |pr_size|		         
  pr_size.value = calculate_size
  # --> another process creates a PullRequestSize record at this point <--
  pr_size.save! # <-- this save fails because of the unique pull_request_id constraint
end
```

The simplest solution seems to split the create_or_update into 2 different operations

```ruby
def update_or_create_pull_request_size
  create_pull_request_size || update_pull_request_size
end
```

where the create_pull_request_size is almost atomic and in the case if fails because the record already exists the update is executed

## Details
- Split `find_or_initialize_by` into an atomic `create` or a `find_by / save!` to avoid concurrency of creation and update
- Refactor - Extract duplicated code in pull_request_size_spec.rb

### DIFF
--- a/app/services/builders/pull_request_size.rb
+++ b/app/services/builders/pull_request_size.rb
@@ -20,8 +20,10 @@ module Builders
 
     def create_pull_request_size
       ::PullRequestSize.create!(pull_request: pull_request, value: pull_request_size_value)
-    rescue ::ActiveRecord::RecordInvalid
-      nil
+    rescue ::ActiveRecord::RecordInvalid => exception
+      if exception.message != 'Validation failed: Pull request has already been taken'
+        raise exception
+      end
     end
 
     def update_pull_request_size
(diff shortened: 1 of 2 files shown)


mrubi-rootstrap:
I'm closing the PR because I was not able to reproduce the error reported in ExceptionHunter yet.

PR #334, opened 2020-09-24
[EM-174] Update pull request size when base branch changes

SandroDamilano:
## What does this PR do?
It updates the value of the PR size metric every time the base branch of the PR changes.

Resolves [#174](https://rootstrap.atlassian.net/browse/EM-174)

### DIFF
--- a/spec/factories/payloads/pull_request.rb
+++ b/spec/factories/payloads/pull_request.rb
@@ -40,6 +40,34 @@ FactoryBot.define do
     end
     initialize_with { attributes.deep_stringify_keys }
 
+    trait :edited do
+      transient do
+        changed { 'title' }
+        from { 'Old PR Title' }
+      end
+
+      action { 'edited' }
+      changes do
+        {
+          changed => {
+            from: from
+          }
+        }
+      end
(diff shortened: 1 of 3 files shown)


GuilleLeopold (approved):

santib (approved):
minimalist 👏

spec/factories/payloads/pull_request.rb
```diff
@@ -40,6 +40,34 @@
...
+    trait :edited do
+      transient do
+        changed { 'title' }
+        from { 'Old PR Title' }
+      end
+
+      action { 'edited' }
+      changes do
+        {
+          changed => {
```
santib, line 52:
```suggestion
          changed: {
```

SandroDamilano: It doesn't work the same way. I want the key to be whatever is in the `changed` variable (`'title'` is the default). If I use `changed:`, the key will be the symbol `:changed`.

santib: you are right, my bad 😄

PR #335, opened 2020-09-24
FEATURE EM176: Pull request is not appearing in external contributions

GuilleLeopold:
Get pull requests to external repositories from the Events API filtering by pull request type in events


.rubocop.yml
```diff
@@ -31,7 +31,7 @@ Lint/AmbiguousBlockAssociation:
 Metrics/AbcSize:
   # The ABC size is a calculated magnitude, so this number can be an Integer or
   # a Float.
-  Max: 15
+  Max: 17
```
santib, line 34: why 17? like did you get the number from somewhere?

app/services/builders/external_pull_request.rb
```diff
@@ -1,16 +1,20 @@
...
-        pull_request.owner_id = User.find_by!(login: @pull_request_data[:user][:login]).id
+      pull_request_payload = @pull_request_event_data.dig(:payload, :pull_request)
+
+      ::ExternalPullRequest
+        .find_or_create_by!(github_id: pull_request_payload[:id]) do |pull_request|
+        pull_request.html_url = pull_request_payload[:html_url]
+        pull_request.body = pull_request_payload[:body]
+        pull_request.title = pull_request_payload[:title]
+        pull_request.owner_id = User.find_by!(login: @pull_request_event_data.dig(:actor, :login))
+                                    .id
```
santib, line 17:
```suggestion
        pull_request.owner_id = User.find_by!(
          login: @pull_request_event_data.dig(:actor, :login)
        ).id
```

app/services/company_member_events_discriminator.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def call
+    @events.reject { |event| company_members.include? event[:repo][:name].split('/').first }
+  end
+
+  private
+
+  def company_members
+    User.pluck(:login) << ENV.fetch('GITHUB_ORGANIZATION', 'rootstrap')
```
santib, line 13: I'd remove the default. Remember this is an OSS project, so I'd like to leave no references to rootstrap

app/services/github_client/user.rb
```diff
@@ -4,11 +4,23 @@ def initialize(username)
...
-      response = connection.get("/users/#{@username}/repos") do |request|
-        request.params['type'] = 'member'
+    def pull_request_events
+      events = (1..10).map do |page|
+        response = connection.get("/users/#{@username}/events/public") do |request|
+          request.params['page'] = page
+        end
+        results = JSON.parse(response.body, symbolize_names: true)
+
+        if results.empty? || older_than_a_day(results.last[:created_at]) || results.size < 30
```
santib, line 14: you added `older_than_a_day` condition because there will be a scheduled task running once per day? What if for some reason it fails one day and doesn't run? What other options do we have? Maybe storing the date of the last run?

app/services/github_client/user.rb
```diff
@@ -4,11 +4,23 @@ def initialize(username)
...
+          break results
+        end
       end
-      JSON.parse(response.body, symbolize_names: true)
+
+      events.flatten.select { |pr| pr[:type] == 'PullRequestEvent' }
```
santib, line 19:
```suggestion
      events.flatten.select { |event| event[:type] == 'PullRequestEvent' }
```

db/migrate/20200611190026_create_departments.rb
```diff
@@ -1,9 +1,6 @@
 class CreateDepartments < ActiveRecord::Migration[6.0]
   def up
-    create_table :departments do |t|
-
-      t.timestamps
-    end
+    create_table :departments, &:timestamps
```
santib, line 3: 😅 this is a pretty weird notation for migrations

db/structure.sql
```diff
@@ -47,7 +47,6 @@ CREATE TYPE public.lang AS ENUM (
     'android',
     'others',
     'unassigned',
-    'vuejs',
```
santib, line 50: ?

app/services/company_member_events_discriminator.rb
```diff
@@ -0,0 +1,15 @@
+class CompanyMemberEventsDiscriminator < BaseService
```
santib, line 1: this file is missing tests 😇

santib (approved):
Nice 🤩

app/services/github_client/user.rb
SandroDamilano: I'm not sure I like the idea of tying this method to that time restriction. If we wanted to do a backfill we'd have to get rid of that, for example. I'd like it to be optional.
But on a second thought, I think we should be fine if we wanted to add that in another PR down the line.

SandroDamilano, line 14: Why that `30`? Is it the default number of results the endpoint returns? Could we extract it to a constant?

app/services/builders/external_project.rb
```diff
@@ -6,9 +6,10 @@ def initialize(repository_data)
...
     def call
       ::ExternalProject.find_or_initialize_by(github_id: @repository_data[:id]) do |project|
-        project.name = @repository_data[:name]
-        project.full_name = @repository_data[:full_name]
-        project.description = @repository_data[:description]
+        name = @repository_data[:name]
+
+        project.name = name.split('/').second
+        project.full_name = name
       end
```
SandroDamilano, line 13: We could now save the project here, right? Because if we call this object we know for sure we want to keep the project. (And we could get rid of `Builders::ExternalPullRequest#built_or_saved_project`)

spec/services/builders/external_projects_spec.rb
```diff
@@ -19,12 +19,13 @@
...
+    let(:full_name) { "github/#{repo_name}" }
+
     let(:repository_data) do
       {
         id: 247_546_341,
-        name: 'ruby_repo',
-        full_name: 'github/ruby_repo',
-        description: 'this is a uby repo'
+        name: full_name
       }
```
SandroDamilano, line 29: I know it probably seems like an overkill, but why not use the factory `gitub_api_client_pull_requests_events_payload`? It would be more precise to me.

spec/services/builders/external_pull_request_spec.rb
```diff
@@ -6,39 +6,59 @@
   context 'when there is already a pull request' do
     let!(:pull_request) { create(:external_pull_request) }
 
-    let(:pull_request_data) do
+    let(:pull_request_event_data) do
```
SandroDamilano, line 9: Again, we could use the factory `gitub_api_client_pull_requests_events_payload` here. Maybe we could add to it a `pull_request` as a transient attribute?

spec/services/github_client/user_spec.rb
```diff
@@ -1,27 +1,27 @@
 require 'rails_helper'
 
 RSpec.describe GithubClient::User do
-  describe '#repositories' do
+  describe '#pull_request_events' do
```
SandroDamilano, line 4: The method got considerably more complex now. Could we add some more tests? (The time restriction is not tested, for example)

spec/support/helpers/github_api_mocker.rb
```diff
@@ -13,7 +13,7 @@ def stub_get_code_owners_file_ok(custom_content_file = '')
   end
 
   def stub_get_repos_from_user(username, payload = [])
```
SandroDamilano, line 15: Are we still using this method somewhere?

SandroDamilano (approved):
Left some comments, but you did a great job! 👏

.rubocop.yml
GuilleLeopold: Hmm, I thought we have changed this numer in the rails-api-base to 17 but I just checked and didnt find anything

GuilleLeopold: In fact we did that last year, Ownable has 17 as an example

app/services/github_client/user.rb
GuilleLeopold: I guess storing the date of the last run may be the best choice. I added this because iterating over the 10 pages every time we run this, takes a lot of time. We have like 80 users and it would imply doing around 800 requests. We may even have some issues with API requests per minute restrictions

db/migrate/20200611190026_create_departments.rb
GuilleLeopold: Hmmmm... I didnt do this

app/services/builders/external_project.rb
```diff
@@ -6,9 +6,10 @@ def initialize(repository_data)
 
     def call
       ::ExternalProject.find_or_initialize_by(github_id: @repository_data[:id]) do |project|
-        project.name = @repository_data[:name]
-        project.full_name = @repository_data[:full_name]
-        project.description = @repository_data[:description]
+        name = @repository_data[:name]
+
+        project.name = name.split('/').second
+        project.full_name = name
```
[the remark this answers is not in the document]
GuilleLeopold: Yes :) :(

app/services/github_client/user.rb
GuilleLeopold: I will move it to a constant. Its the default (and unchangable) per page value

PR #337, opened 2020-09-29
EM175: add opened_at to external pull request model

GuilleLeopold:
https://rootstrap.atlassian.net/jira/software/projects/EM/boards/42?selectedIssue=EM-175

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -47,7 +47,7 @@ CREATE TYPE public.lang AS ENUM (
     'android',
     'others',
     'unassigned',
-    'react_native'
+    'vuejs'
 );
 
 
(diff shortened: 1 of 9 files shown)


santib (approved):

db/structure.sql
```diff
@@ -47,7 +47,7 @@ CREATE TYPE public.lang AS ENUM (
     'android',
     'others',
     'unassigned',
-    'react_native'
+    'vuejs'
```
SandroDamilano, line 50: Why this change?

SandroDamilano (approved):
🥇

db/structure.sql
GuilleLeopold: It seemed like it was deleted in the previous PR, dont know why. @santib commented on that on the previous PR. And the order changed probably is because postgres version or something like that

GuilleLeopold: i will check, it happened the same on the other PR

PR #340, opened 2020-10-02
add missing tests for company memeber discriminator

GuilleLeopold:


spec/services/company_member_events_discriminator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    let(:event_repo_external) do
+      { repo: { id: 3, name: 'guilleleop/project3' } }
+    end
+
+    let(:events) { [event_repo_organization, event_repo_user, event_repo_external] }
+
+    before do
+      ENV['GITHUB_ORGANIZATION'] = 'mycompany'
```
santib, line 22: maybe you can put this in the `.env.test` ? It's not a good idea to modify ENV VARs in tests because the following ones will be affected by this change, and at some point we might add tests that uses this ENV VAR causing intermittent failures.

Other option would be to stub the env var https://relishapp.com/rspec/rspec-mocks/v/2-11/docs/stubbing-constants

santib (approved):

SandroDamilano (approved):

spec/services/company_member_events_discriminator_spec.rb
SandroDamilano: I usually use `stub_env`, which is very handy.
Here it would be `stub_env('GITHUB_ORGANIZATION', 'mycompany')`

PR #344, opened 2020-10-07
Hotfix/external contributions wrong users

GuilleLeopold:


app/services/github_client/user.rb
```diff
@@ -7,10 +7,25 @@ def initialize(username)
     end
 
     def pull_request_events
-      events = (1..10).map do |page|
+      events = retrieve_events
+
+      events.flatten.select { |event| event[:type] == 'PullRequestEvent' }
+    rescue URI::InvalidURIError => exception
+      ExceptionHunter.track(exception)
```
SandroDamilano, line 14: Would it make sense to return an empty list instead of `nil` here? (Or whatever Exception Hunter returns)
I think that way we wouldn't need to add [this line](https://github.com/rootstrap/rs-code-review-metrics/pull/344/files#diff-f5e889280495856ba8b2fd80796cc080R9).

SandroDamilano (approved):
Looks good!
I think it would be great to have this code on `develop` as well. What would be best, to open another PR pointing to develop or to point this one to develop and making a release immediately after it's merged? I personally prefer the second one, but I'll leave it up to you.
Also this is a reminder to open a task to add a test to this case 😁

app/services/github_client/user.rb
```diff
@@ -7,10 +7,25 @@ def initialize(username)
...
+    end
+
+    private
+
+    def older_than(created_at)
+      DateTime.parse(created_at) < ENV.fetch('SCHEDULE_EXTERNAL_PULL_REQUESTS', 1).to_i.days.ago
+    end
+
+    def retrieve_events
+      (1..10).map do |page|
```
horacio, line 24: WDYT about using `flat_map` here, so we can just do `events.select { |event| event[:type] == 'PullRequestEvent' }` in `#pull_request_events`?

app/services/processors/external/pull_requests.rb
```diff
@@ -6,6 +6,8 @@ def initialize(username)
       end
 
       def call
+        return unless events
```
horacio, line 9:
```suggestion
        return unless events.present?
```

horacio (approved):
👏

app/services/github_client/user.rb
GuilleLeopold: Yes. It seems like ExceptionHunter returns nil. I thought about doing that too. It might be fine

GuilleLeopold:
I was talking with Martin and i think the best is to merge master into develop after merging this one

PR #348, opened 2020-10-13
break does not accumulate results

GuilleLeopold:
This PR fix the issue where the pull_contributions_events doesn't accumulate events on the `map` if the `break` line is executed

### DIFF
--- a/app/services/github_client/user.rb
+++ b/app/services/github_client/user.rb
@@ -9,7 +9,7 @@ module GithubClient
     def pull_request_events
       events = retrieve_events
 
-      events.select { |event| event[:type] == 'PullRequestEvent' }
+      events.flatten.select { |event| event[:type] == 'PullRequestEvent' }
     rescue URI::InvalidURIError => exception
       ExceptionHunter.track(exception)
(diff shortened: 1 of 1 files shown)


app/services/github_client/user.rb
```diff
@@ -18,22 +18,21 @@ def pull_request_events
     private
 
     def older_than(created_at)
-      DateTime.parse(created_at) < ENV.fetch('SCHEDULE_EXTERNAL_PULL_REQUESTS', 1).to_i.days.ago
+      DateTime.parse(created_at) < ENV.fetch('SCHEDULE_EXTERNAL_PULL_REQUESTS', 2).to_i.days.ago
```
GuilleLeopold, line 21: I changed this to 2 as default to make sure we are not missing events

santiagovidal (approved):

martinjaimem (approved):
tests 🙏

GuilleLeopold:
@martinjaimem There's a ticket that we already created for that, yes. We are aware that we have a debt on testing this particular method

PR #164, opened 2020-05-12
Release/2020.05.12

hdamico:
## Release

- Merge time metric.
- New Cron Jobs.
- RequestReview creation fix.
- Daily and Weekly metrics.

PR #351, opened 2020-10-30
[EM-164] Add external contributions through admin

SandroDamilano:
## What does this PR do?
It adds External Pull Requests to the admin to be able to create new ones manually

Screen-Recording-2020-10-30-at-1

Resolves [#164](https://rootstrap.atlassian.net/browse/EM-164)


.env.example
```diff
@@ -5,9 +5,23 @@ GITHUB_ANALYZER_PASSWORD=postgres
...
 
+# Own Github organization
+GITHUB_ORGANIZATION=rootstrap
+
+# Worpress data
+WORDPRESS_SITE_ID=
+WORDPRESS_CLIENT_ID=
+WORDPRESS_CLIENT_SECRET=
+WORDPRESS_USERNAME=
+WORDPRESS_PASSWORD=
```
horacio, line 16: Are these variables related to this particular PR?

app/models/external_pull_request.rb
```diff
@@ -26,6 +27,8 @@
 
 class ExternalPullRequest < ApplicationRecord
   belongs_to :external_project
+  alias project external_project
```
horacio, line 30: I think this would work for read-only ops, but what about assignments? Please, disregard this if it isn't something of importance. I'm bringing it up just in case.

db/structure.sql
```diff
@@ -572,6 +573,7 @@ CREATE TABLE public.external_pull_requests (
     external_project_id bigint NOT NULL,
     created_at timestamp(6) without time zone NOT NULL,
     updated_at timestamp(6) without time zone NOT NULL,
+    number integer,
```
horacio, line 576: Just out of curiosity: why aren't we adding a NOT NULL constraint here?

horacio (approved):
Looking good, Sandro! 👏  I left some non-blocking comments, mostly doubts. I checked out this branch and tried locally, works like a charm.

santiagovidal (approved):

.env.example
SandroDamilano: No, they are not! I think I originally added these for when someone new would join the team, so that they could have the correct env vars. But well, I finished a bit too late 😅 
In any case, I could totally add these changes in another PR

app/models/external_pull_request.rb
SandroDamilano: Yeah, it probably doesn't work for assignments, but I just needed it for read operations anyway (to make `ExternalPullRequest` polymorphic with `PullRequest` for the `project` message).
Do you think it would be best for it to work for assignments as well?

db/structure.sql
SandroDamilano: Just so that we didn't need to add a default value for old records (this way, old records will have the `number` attribute in `nil` and not some random number that could be misleading).
If we made a backfill, then we could add the constraint

.env.example
horacio: Personally I think it's fine

app/models/external_pull_request.rb
horacio: If it isn't something in our short-term backlog, then I would leave it as it is and tackle it when it becomes a problem.

db/structure.sql
horacio: Ahhhh I see. Nice!

nicofh (approved):

PR #352, opened 2020-11-02
Add state to external pr

nicofh:
## What does this PR do?
Adds state attribute to external_pull_requests and shows it on External Contributors page

Resolves https://rootstrap.atlassian.net/browse/EM-177

### DIFF
--- a/app/services/builders/external_pull_request.rb
+++ b/app/services/builders/external_pull_request.rb
@@ -8,17 +8,30 @@ module Builders
     def call
       pull_request_payload = @pull_request_event_data.dig(:payload, :pull_request)
 
-      ::ExternalPullRequest
-        .find_or_create_by!(github_id: pull_request_payload[:id]) do |pull_request|
-        pull_request.html_url = pull_request_payload[:html_url]
-        pull_request.body = pull_request_payload[:body]
-        pull_request.title = pull_request_payload[:title]
-        pull_request.opened_at = pull_request_payload[:created_at]
-        pull_request.owner_id = User.find_by!(
+      pull_request = ::ExternalPullRequest
+                     .find_or_initialize_by(
+                       github_id: pull_request_payload[:id]
+                     ) do |ext_pull_request|
+        ext_pull_request.html_url = pull_request_payload[:html_url]
+        ext_pull_request.body = pull_request_payload[:body]
+        ext_pull_request.title = pull_request_payload[:title]
+        ext_pull_request.opened_at = pull_request_payload[:created_at]
+        ext_pull_request.owner_id = User.find_by!(
           login: @pull_request_event_data.dig(:actor, :login)
         ).id
-        pull_request.external_project_id = @external_project.id
+        ext_pull_request.external_project_id = @external_project.id
       end
+      pull_request.state = assign_state(pull_request_payload)
+      pull_request.save!
+      pull_request
+    end
+
+    private
+
+    def assign_state(pull_request_payload)
+      return 'merged' if pull_request_payload[:merged]
+
+      pull_request_payload[:state]
     end
   end
 end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -35,6 +35,17 @@ CREATE TYPE public.department_name AS ENUM (
 
 
 --
+-- Name: external_pull_request_state; Type: TYPE; Schema: public; Owner: -
+--
+
+CREATE TYPE public.external_pull_request_state AS ENUM (
+    'open',
+    'closed',
+    'merged'
+);
+
+
+--
 -- Name: lang; Type: TYPE; Schema: public; Owner: -
 --
 
@@ -47,7 +58,8 @@ CREATE TYPE public.lang AS ENUM (
     'android',
     'others',
     'unassigned',
-    'vuejs'
+    'vuejs',
+    'react_native'
 );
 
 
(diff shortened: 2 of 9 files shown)


app/services/builders/external_pull_request.rb
```diff
@@ -8,17 +8,27 @@ def initialize(pull_request_event_data, external_project)
...
+        ext_pull_request.body = pull_request_payload[:body]
+        ext_pull_request.title = pull_request_payload[:title]
+        ext_pull_request.opened_at = pull_request_payload[:created_at]
+        ext_pull_request.owner_id = User.find_by!(
           login: @pull_request_event_data.dig(:actor, :login)
         ).id
-        pull_request.external_project_id = @external_project.id
+        ext_pull_request.external_project_id = @external_project.id
       end
+      pull_request.update!(state: assign_state(pull_request_payload))
```
SandroDamilano, line 22: To avoid accessing the DB twice, we could just do a `find_or_initialize_by` instead of a `find_or_create_by!`

SandroDamilano, line 22: I'd add at least a test for this change

db/structure.sql
```diff
@@ -132,6 +133,8 @@ CREATE TYPE public.review_state AS ENUM (
 
 SET default_tablespace = '';
 
+SET default_table_access_method = heap;
```
SandroDamilano, line 136: What is this change for?

spec/services/processors/external/pull_requests_spec.rb
```diff
@@ -22,4 +22,18 @@
         .to change { ExternalPullRequest.count }.by(1)
     end
   end
+
+  context 'when user has an open pull request in a given project which gets merged' do
+    let!(:external_pull_request) do
+      ExternalPullRequest.destroy_all
```
SandroDamilano, line 28: Why are we doing this?

db/migrate/20201029160139_add_state_to_external_pull_request.rb
```diff
@@ -0,0 +1,5 @@
+class AddStateToExternalPullRequest < ActiveRecord::Migration[6.0]
+  def change
+    add_column :external_pull_requests, :state, :string
```
SandroDamilano, line 3: Instead of it being just a string, we could add a type to the DB and an enum to the code, like we did with `Events::PullRequest` or `Metric`

SandroDamilano (commented):
Good job! I just left a few comments.
Also, be careful with the addition of the files `dump.rdb` and `package-lock.json` (maybe the latter should be versioned, actually, but probably not the former)

db/structure.sql
nicofh: I didn't manually make this. I don't know what it is or how it got set 😕.

spec/services/processors/external/pull_requests_spec.rb
nicofh: My mistake I was trying something here. Will delete this as it's not necessary.

db/structure.sql
horacio: @nicofh what PostgreSQL version are you on?

nicofh: 13.0

horacio: Cool, thanks. I don't think this should pose any problem, but in any case we can disregard `structure.sql` diffs not related to our changesets.

horacio (approved):
Looking good!

app/services/builders/external_pull_request.rb
SandroDamilano: In that case you're right, but if the record is not present, we would hit the DB twice, first with the `create` and then with the `update`. If you use `find_or_initialize_by`, the DB will be hit just once.
Or am I missing something?

SandroDamilano: Is there any reason why not adding specs to `spec/services/builders/external_pull_request_spec.rb`?
If the change is in that file, I find it best to have it tested in the own class spec file (which doesn't mean testing it where you did is wrong, in fact it's a good idea to have both specs).

db/structure.sql
SandroDamilano: I've experienced some problems with this line before, because it doesn't work with versions of postgres earlier than 12. We may still not be interested in avoiding this problem, but I think that if we don't know what this line is doing, we may want to keep it away from the file.

db/migrate/20201102163702_add_state_to_external_pull_request.rb
```diff
@@ -0,0 +1,9 @@
+class AddStateToExternalPullRequest < ActiveRecord::Migration[6.0]
+  def change
+    execute <<-SQL
+      CREATE TYPE state AS ENUM ('open', 'closed', 'merged');
+    SQL
```
SandroDamilano, line 5: Be careful, because this SQL command is not reversible. You can check out this migration to see how to make it reversible `db/migrate/20200305150445_change_review_requests_status.rb`

db/structure.sql
```diff
@@ -2227,6 +2242,9 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20200908173642'),
 ('20200911143301'),
 ('20200922131539'),
-('20200929205630');
+('20200929205630'),
+('20201029160139'),
+('20201102155621'),
+('20201102163702');
```
SandroDamilano, line 2248: I think you added more migrations here than what you should have. I think we can keep just the last one, `20201102163702`

app/services/builders/external_pull_request.rb
nicofh: I messed up the files, will move this to `spec/services/builders/external_pull_request_spec.rb`. thanks

nicofh: you are right

SandroDamilano (approved):
Great work!

PR #356, opened 2020-11-11
removed wordpress get blog endpoint call that was not working

nicofh:
## What does this PR do?
Tech blog posts count was 0 for some months which have blog posts. That was because BlogPost were loaded without technologies because of an external endpoint failure. We already had that info so that call was removed.

Resolves https://rootstrap.atlassian.net/browse/EM-186

### DIFF
--- a/app/services/processors/blog_posts_updater.rb
+++ b/app/services/processors/blog_posts_updater.rb
@@ -7,7 +7,9 @@ module Processors
         blog_post.slug = blog_post_payload[:slug]
         blog_post.status = blog_post_payload[:status]
         blog_post.url = blog_post_payload[:URL]
-        blog_post.technologies = technologies_for(blog_post)
+        if blog_post.technologies.empty?
+          blog_post.technologies = technologies_for(blog_post_payload)
+        end
 
         blog_post.save!
       end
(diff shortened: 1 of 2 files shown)


hdamico (approved):

SandroDamilano (approved):
Good job!

horacio (approved):

app/services/wordpress_service.rb
```diff
@@ -24,10 +24,6 @@ def blog_post_views(blog_post_id)
     get_response_body("sites/#{site_id}/stats/post/#{blog_post_id}")
   end
 
-  def blog_post(blog_post_id)
-    get_response_body("sites/#{site_id}/posts/#{blog_post_id}")
-  end
-
```
SandroDamilano, line 30: I know we are going to stop using it, but I'd keep this method regardless, just in case we (or they) can fix the problem sometime down the line.
Is there any linter forcing you to do it?

nicofh: code climate coverage is failing because is under 99%, I thought it was because of this but it seems not

SandroDamilano: Could you restore it, then?

PR #360, opened 2020-11-13
fix external pr status

nicofh:
## What does this PR do?
Fixes external pr status bug. It adds a get request to pull last status.


martinjaimem (approved):

SandroDamilano (approved):
Good job!

Just a comment, do we need the changes in `spec/services/builders/external_pull_request_spec.rb` with this new approach?

PR #367, opened 2020-11-24
Update Faraday gem version and improve error data sent to Exception Hunter

SandroDamilano:
## What does this PR do?
It improves the info logged in Exception Hunter about HTTP errors

image

To do so, Faraday had to be updated to the latest release

Resolves ¯\_(ツ)_/¯


app/services/base_service.rb
```diff
@@ -2,6 +2,8 @@
 
 # Base class for service classes
 class BaseService
+  include RequestErrorTrackingHelper
```
santib, line 5: `Helper`s are specifically for views in the context of rails apps, I don't like having them being included in services. Maybe you can move the `RequestErrorTrackingHelper` inside the `services` folder

santib (approved):
Left one comment, approved once it's resolved

app/services/base_service.rb
SandroDamilano: Great, thank you for the insight!

nicofh (approved):

PR #368, opened 2020-11-24
Projects without CC

sebastiancaraballo:
## What does this PR do?
- Added `CodeClimate::ProjectsWithoutCC` in order to return projects which do not have CC data for the selected time range. Added `See project` button for each project in order to open its metrics page.
- Changed styles to existent CC report, ordered them by name and added `See project` button as well.
- Added test.

### Preview:
ezgif com-gif-maker (3)

Resolves [See list of projects that don't have CodeClimate in CodeClimate report](https://rootstrap.atlassian.net/browse/EM-191?atlOrigin=eyJpIjoiZjk0ZWNlYTgyMjU3NDJmOGE4OGRjM2VlYzYzNmY3NzkiLCJwIjoiaiJ9)


app/models/project.rb
```diff
@@ -80,6 +80,19 @@ class Project < ApplicationRecord
     joins(:pull_requests).where('pull_requests.merged_at > ?', date)
   }
 
+  scope :by_department, lambda { |department|
+    joins(language: :department).where(departments: { id: department.id })
+  }
+
+  scope :by_metrics_time, lambda { |from|
+    joins(:code_climate_project_metric)
```
nicofh, line 88: if you want projects without cc, then does it make sense to join code_climate_project_metric table?

app/services/code_climate/projects_without_c_c.rb
```diff
@@ -0,0 +1,43 @@
...
+    attr_reader :department, :from, :languages
+
+    def initialize(department:, from:, languages:)
+      @department = department
+      @from = from.to_i
+      @languages = languages
+    end
+
+    def call
+      projects_in_given_languages(projects_in_department)
```
nicofh, line 12: Are you returning no cc projects within the time range?

nicofh (approved):

app/services/code_climate/projects_without_c_c.rb
```diff
@@ -0,0 +1,43 @@
+module CodeClimate
+  class ProjectsWithoutCC < BaseService
```
SandroDamilano, line 2: I'd like this service to be properly tested in its own spec file covering edge cases

app/services/code_climate/projects_details_service.rb
```diff
@@ -9,8 +9,6 @@ def initialize(department:, from:, technologies: [])
     end
 
     def call
-      return ProjectsSummary.new unless code_climate_metrics.any?
```
SandroDamilano, line 12: Nice job getting rid of this line 😁

SandroDamilano (approved):
I'd like a little more testing, but good job overall!

app/services/code_climate/projects_without_c_c.rb
sebastiancaraballo: Based on the current implementation at `CodeClimate::ProjectsDetailsService`, it filters projects by the metrics' `snapshot_time` range, so the same happens here. So, when no-cc projects are returned, it's because their metrics are not within the `snapshot_time: from.weeks.ago..Time.zone.now` range.

app/services/code_climate/projects_without_c_c.rb
sebastiancaraballo: Will address on next PR.

sebastiancaraballo: Will address on next PR.

PR #370, opened 2020-11-30
show projects without cc and refactor cc summary

nicofh:
## What does this PR do?
Refactor code climate summary so it's consistent with overview numbers in `Department Metrics`
Also adds `projects without code climate` data.

Resolves https://rootstrap.atlassian.net/jira/software/projects/EM/boards/42?selectedIssue=EM-187


app/services/code_climate/projects_summary_service.rb
```diff
@@ -58,16 +68,38 @@ def metrics_in_given_technologies
...
         metrics_in_department
       end
     end
 
     def metrics_in_department
-      CodeClimateProjectMetric
-        .joins(project: { language: :department })
-        .where(departments: { id: department.id })
+      Project.joins(:code_climate_project_metric)
+             .joins(:language)
```
hdamico, line 79: Cant you do this? `Project.joins(:code_climate_project_metric, :language)`

GuilleLeopold (approved):

hdamico (approved):

app/models/project.rb
```diff
@@ -91,6 +91,17 @@ class Project < ApplicationRecord
...
+  scope :with_no_cc, lambda {
+    left_joins(:code_climate_project_metric).where(code_climate_project_metrics: { id: nil })
+  }
+
+  scope :with_no_cc_rate, lambda {
+    left_joins(:code_climate_project_metric)
+      .where(code_climate_project_metrics: { code_climate_rate: nil })
+  }
+
+  scope :with_no_cc_nor_cc_rate, lambda {
```
santib, line 103: I might be wrong but I interpret `nor` in `with_no_cc_nor_cc_rate` as "projects without cc AND without cc rate", and `with_no_cc_or_cc_rate` as "projects without cc OR without cc rate",

app/models/project.rb
```diff
@@ -91,6 +91,17 @@ class Project < ApplicationRecord
 
   scope :by_language, lambda { |languages|
     joins(:language).where(languages: { name: languages })
+  scope :with_no_cc, lambda {
```
santib, line 94: maybe `without` reads better than `with_no` and is also more consistent with some other places

app/services/code_climate/projects_summary_service.rb
```diff
@@ -58,16 +68,38 @@ def metrics_in_given_technologies
...
+             .distinct
+             .relevant
+    end
+
+    def projects_without_cc_count
+      (ids_without_rate - ids_with_rate).count
+    end
+
+    def ids_without_rate
+      projects_without_cc_count = Project.with_no_cc_nor_cc_rate
```
santib, line 90: names look a bit confusing, you are using the same name for this variable as for the method above, but both yield different results, right?

app/models/project.rb
nicofh: This was the only way I could make the OR work for scopes. I can change it to `without_cc_or_cc_rate`

app/services/code_climate/projects_summary_service.rb
```diff
@@ -38,12 +39,21 @@ def open_issues_count_average
...
-      code_climate_ratings = code_climate_metrics.pluck(:code_climate_rate).compact
-      code_climate_ratings.each_with_object(Hash.new(0)) do |code_climate_rate, ratings|
-        ratings[code_climate_rate] += 1
+      @code_climate_ratings = code_climate_metrics.pluck(:id,
+                                                         :code_climate_rate,
+                                                         'code_climate_project_metrics.created_at')
+      @code_climate_ratings = sort_ratings_by_date
+      @code_climate_ratings.each_with_object(Hash.new(0)) do |code_climate_metric, ratings|
+        letter = code_climate_metric[1]
+        ratings[letter] += 1 if letter.present?
```
santib, line 48:
```suggestion
      @ratings ||= begin
        sort_ratings_by_date(
          code_climate_metrics.pluck(:id, :code_climate_rate, 'code_climate_project_metrics.created_at')
        ).each_with_object(Hash.new(0)) do |code_climate_metric, ratings|
          letter = code_climate_metric[1]
          ratings[letter] += 1 if letter.present?
        end
```

I think something like that would look simpler to me

santib: notice that I pass the intermediate result as a param to `sort_ratings_by_date` and that you can call `ratings` directly here https://github.com/rootstrap/rs-code-review-metrics/pull/370/files#diff-229e40836c3086defd645ce882813e3ee45dcce4d1b88f760fc1000c72454cdfR102

app/services/code_climate/projects_summary_service.rb
```diff
@@ -38,12 +39,21 @@ def open_issues_count_average
...
       end
     end
 
+    def sort_ratings_by_date
+      @code_climate_ratings.sort_by { |id_rate_date| [id_rate_date[0], id_rate_date[2]] }
+                           .each_with_object({}) { |id_rate, hsh| hsh[id_rate[0]] = id_rate[1] }
```
santib, line 54: what is this doing? the method is called `sort_...` but the sort was already done in the previous line

app/services/code_climate/projects_summary_service.rb
```diff
@@ -38,12 +39,21 @@ def open_issues_count_average
     end
 
     def ratings
-      code_climate_ratings = code_climate_metrics.pluck(:code_climate_rate).compact
-      code_climate_ratings.each_with_object(Hash.new(0)) do |code_climate_rate, ratings|
-        ratings[code_climate_rate] += 1
+      @code_climate_ratings = code_climate_metrics.pluck(:id,
```
santib, line 42: I'm wondering if it might be better to use `select` instead of `pluck` so that we can work with hashes instead of arrays that make it hard to understand what is in `result[0]` for example. On the other hand I think `pluck` is more performant.. but in this case clarity might be more important than performance 🤔 thoughts?

app/services/code_climate/projects_summary_service.rb
nicofh: my bad, this code was formerly on the method above, will change the names, thanks

app/services/code_climate/projects_summary_service.rb
nicofh: I want to keep only the last rate per project, so after the sorting by date, I store the last value for each project.

app/services/code_climate/projects_summary_service.rb
nicofh: whenever using arrays i tried to use a nomeclature like `id_rate_date` so that `id` is [0], `rate` [1] and `date` [2]

app/services/code_climate/projects_summary_service.rb
nicofh: the thing is that in that line I need a hash with [id,rate] and `ratings` returns [rate,count]. I just refactored it to make it simpler, tell me what you think.

santib (approved):
Good job 👏

PR #374, opened 2020-12-07
Change filters

sebastiancaraballo:
## What does this PR do?
- This PR unifies CodeClimate metrics. Now, the week selector filters projects by their activity and relevance, instead of the metrics snapshot time.

Resolves [#EM-195](https://rootstrap.atlassian.net/browse/EM-195?atlOrigin=eyJpIjoiZDUwODkzNzUzOTZmNDZmMDlkMjNhZDJmZGI2ODQzZjkiLCJwIjoiaiJ9)

### DIFF
--- a/app/services/code_climate/projects_details_service.rb
+++ b/app/services/code_climate/projects_details_service.rb
@@ -42,7 +42,9 @@ module CodeClimate
 
     def metrics_in_time_period
       if from.positive?
-        metrics_in_department.where(snapshot_time: from.weeks.ago..Time.zone.now)
+        metrics_in_department.merge(Project.with_activity_after(from.weeks.ago))
+                             .group('code_climate_project_metrics.id')
+                             .group('projects.id')
       else
         metrics_in_department
       end
@@ -50,6 +52,7 @@ module CodeClimate
 
     def metrics_in_department
       CodeClimateProjectMetric
+        .with_rates
         .joins(project: { language: :department })
         .where(departments: { id: department.id })
         .order('LOWER(projects.name)')
--- a/app/services/code_climate/projects_without_c_c.rb
+++ b/app/services/code_climate/projects_without_c_c.rb
@@ -9,28 +9,26 @@ module CodeClimate
     end
 
     def call
-      projects_in_given_languages(projects_in_department)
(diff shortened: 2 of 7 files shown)


app/services/code_climate/projects_details_service.rb
```diff
@@ -42,14 +42,17 @@ def metrics_in_given_technologies
 
     def metrics_in_time_period
       if from.positive?
-        metrics_in_department.where(snapshot_time: from.weeks.ago..Time.zone.now)
+        metrics_in_department.merge(Project.with_activity_after(from.weeks.ago))
```
santib, line 45: 💯 😉

app/services/code_climate/projects_without_c_c.rb
```diff
@@ -9,28 +9,22 @@ def initialize(department:, from:, languages:)
     end
 
     def call
-      projects_in_given_languages(projects_in_department)
-        .where.not(id: projects_with_cc_ids)
-        .order('LOWER(projects.name)')
```
santib, line 14: why isn't the order needed anymore?

spec/services/code_climate/projects_without_c_c_spec.rb
```diff
@@ -35,9 +59,17 @@
                snapshot_time: Time.zone.now.ago(10.weeks)
       end
 
-      it 'returns project' do
-        projects = subject
-        expect(projects.first).to eq project
+      context 'and the project activity does not match' do
+        let(:project) { create :project, :with_activity, :internal, language: ruby_lang }
+        let(:period) { 2 }
```
santib, line 64: The problem I see here is that this test depends on a factory to set a specific attributes so that `the project activity does not match`. I'm ok with the trait `with_activity` but I'd like this place to be clear that the activity it's setting is older than the `period`

santib (approved):
Good job 👏

app/services/code_climate/projects_without_c_c.rb
sebastiancaraballo: oops

SandroDamilano (approved):
Looks good! 💯

PR #375, opened 2020-12-08
Add Time to 2nd review distribution graph

sebastiancaraballo:
## What does this PR do?
- Add Time to 2nd review distribution graph.
- Refactored & added tests.

Preview:
ezgif com-gif-maker

Resolves [#EM-154](https://rootstrap.atlassian.net/browse/EM-154?atlOrigin=eyJpIjoiYmI1NzhhMjEyNmUxNDRjMTlhZDdiM2FhNzY3OWRiZjkiLCJwIjoiaiJ9)

### DIFF
--- a/app/services/builders/chartkick/project_distribution_data.rb
+++ b/app/services/builders/chartkick/project_distribution_data.rb
@@ -10,17 +10,19 @@ module Builders
       private
 
       def retrieve_records
-        merge_times
+        metric.retrieve_records(entity_id: @entity_id, time_range: @query[:value_timestamp])
       end
 
-      def merge_times
-        ::MergeTime.joins(pull_request: :project)
-                   .where(projects: { id: @entity_id })
-                   .where(pull_requests: { merged_at: @query[:value_timestamp] })
+      def metric_name
+        @query[:name]
+      end
+
+      def metric
+        @metric ||= ProjectDistributionDataMetrics.const_get(metric_name.to_s.camelize).new
       end
 
       def resolve_interval(entity)
-        Metrics::IntervalResolver::Time.call(entity.value_as_hours)
+        metric.resolve_interval(entity)
       end
     end
   end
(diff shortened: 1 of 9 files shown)


SandroDamilano (approved):
Nice work! 👏

app/services/builders/chartkick/project_distribution_data.rb
```diff
@@ -10,17 +10,19 @@ def call
...
-      def merge_times
-        ::MergeTime.joins(pull_request: :project)
-                   .where(projects: { id: @entity_id })
-                   .where(pull_requests: { merged_at: @query[:value_timestamp] })
+      def metric_name
+        @query[:name]
+      end
+
+      def metric
+        @metric ||= ProjectDistributionDataMetrics.const_get(metric_name.to_s.camelize).new
```
santib, line 21: The cons I see with this dynamic approach is that it'd be easy to introduce a bug in the future. Another possible approach would be to have

```ruby
def metric
  @metric ||= case @query[:name]
              when :merge_time
                ProjectDistributionDataMetrics::MergeTime.new
              when :review_turnaround
                ProjectDistributionDataMetrics::ReviewTurnaround.new
              end
end
```

but it's true it's more verbose. So yeah.. I'm ok either way

app/services/builders/chartkick/project_distribution_data_metrics/review_turnaround.rb
```diff
@@ -0,0 +1,18 @@
+module Builders
+  module Chartkick
+    module ProjectDistributionDataMetrics
+      class ReviewTurnaround
+        def retrieve_records(entity_id:, time_range:)
+          ::CompletedReviewTurnaround
+            .joins(review_request: :project)
+            .where(projects: { id: entity_id })
+            .where(created_at: time_range)
```
santib, line 9: don't we have other attribute more accurate than the `created_at` that's just the creation time of the record in our DB?

santib (approved):
Good work

app/services/builders/chartkick/project_distribution_data.rb
sebastiancaraballo: I'll keep it in mind.

app/services/builders/chartkick/project_distribution_data_metrics/review_turnaround.rb
sebastiancaraballo: I don't think we do.

PR #391, opened 2020-12-17
Enabled features settings

sebastiancaraballo:
## What does this PR do?
- This PR aims to hide the following sections:
1. Users section at Development Metrics sidebar.
2. Department metrics per-technology detail graph.
3. Project metrics per-user detail graphs.
4. Project metrics codeowners section.

- Create `SettingsService` in order to manage all different settings and refactor existent code.

Resolves [#EM-205](https://rootstrap.atlassian.net/browse/EM-205?atlOrigin=eyJpIjoiZjIwNmIwOWVkMjNjNDk0YmE0YWRjNjQzNGEzNWZkNzYiLCJwIjoiaiJ9) [#EM-206](https://rootstrap.atlassian.net/browse/EM-206?atlOrigin=eyJpIjoiYzcyOGE0YWE1OGI2NDM1YmIxZmQ4M2VlZDk2ZjA2MmUiLCJwIjoiaiJ9) [#EM-207](https://rootstrap.atlassian.net/browse/EM-207?atlOrigin=eyJpIjoiYmFkNDhhNGFiMzk1NDBiZGFhMmU2ZmU4NzgyZWZhZjgiLCJwIjoiaiJ9) [#EM-208](https://rootstrap.atlassian.net/browse/EM-208?atlOrigin=eyJpIjoiOWVkMTlhZDkyZjEwNDBiOTk2YTYyMGY1ODkyYjViYTgiLCJwIjoiaiJ9)

### DIFF
--- a/app/controllers/development_metrics_controller.rb
+++ b/app/controllers/development_metrics_controller.rb
@@ -1,5 +1,6 @@
 class DevelopmentMetricsController < ApplicationController
   layout 'sidebar_metrics'
+  include LoadSettings
 
   def index; end
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,15 @@ This project comes with:
 
 - In order to change Department Success Rates time limit setting, create a new `Setting` with key prefix `success_rate`, followed by the department name and the metric name.
 - Example: `Setting.create!(key: 'success_rate_backend_merge_time', value: '12')`
+Possible values: 12 | 24 (default) | 36 | 48 | 60 | 72
+
+### Enabled Features Settings
+Possible values: true | false (default)
+
+- `enabled_users_section` enables Users section at Development Metrics sidebar.
+- `enabled_department_per_tech_graph` enables Department per technology detail graph.
+- `enabled_project_codeowners_section` enables Project codeowners section.
+- `enabled_project_per_user_graph` enables Project per-user detail graphs.
 
 ## Code quality
 
--- a/app/controllers/code_climate/departments_controller.rb
+++ b/app/controllers/code_climate/departments_controller.rb
@@ -1,6 +1,7 @@
 module CodeClimate
   class DepartmentsController < ApplicationController
     layout 'sidebar_metrics'
+    include LoadSettings
 
     def show
       @code_climate = code_climate_department_summary
--- a/app/controllers/concerns/load_settings.rb
+++ b/app/controllers/concerns/load_settings.rb
@@ -1,0 +1,32 @@
+module LoadSettings
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :load_settings
+  end
+
+  private
+
+  def load_settings
+    enabled_users_section
+    enabled_department_per_tech_graph
+    enabled_project_codeowners_section
+    enabled_project_per_user_graph
+  end
+
+  def enabled_users_section
+    @enabled_users_section ||= SettingsService.enabled_users_section
+  end
+
+  def enabled_department_per_tech_graph
+    @enabled_department_per_tech_graph ||= SettingsService.enabled_department_per_tech_graph
+  end
+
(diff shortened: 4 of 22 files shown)


app/controllers/concerns/concerns/load_settings.rb
```diff
@@ -0,0 +1,32 @@
+module Concerns
```
santib, line 1: why `concerns/concerns` ? I think that concerns aren't usually put under a `Concerns` module

app/controllers/development_metrics_controller.rb
```diff
@@ -1,5 +1,6 @@
 class DevelopmentMetricsController < ApplicationController
   layout 'sidebar_metrics'
+  include Concerns::LoadSettings
```
santib, line 3: is this concerns going to be reused? Or why did you use a concern and not just put the logic in this controller?

config/rails_best_practices.yml
```diff
@@ -34,7 +34,9 @@ RemoveUnusedMethodsInModelsCheck: { except_methods: [
     'ReviewTurnaround#build'
   ] }
 ReplaceComplexCreationWithFactoryMethodCheck: { attribute_assignment_count: 2 }
-ReplaceInstanceVariableWithLocalVariableCheck: { }
+ReplaceInstanceVariableWithLocalVariableCheck: { ignored_files: [
+  'views/development_metrics/_sidebar.erb'
```
santib, line 38: I'd rather fix it instead of ignoring it. Partials shouldn't use instance variables, you can pass them via locales. https://andycroll.com/ruby/dont-use-instance-variables-in-partials/

santib (approved):
Good job. Approved after comments are resolved

app/controllers/concerns/concerns/load_settings.rb
```diff
@@ -0,0 +1,32 @@
...
+  module LoadSettings
+    extend ActiveSupport::Concern
+
+    included do
+      before_action :load_settings
+    end
+
+    def load_settings
```
santir489, line 9: should this be private ?

horacio (approved):

app/controllers/concerns/concerns/load_settings.rb
sebastiancaraballo: yes

app/controllers/development_metrics_controller.rb
sebastiancaraballo: Since there may appear more settings in the future, I think it's a good idea to place this on a concern. I don't quite like this on the controller.

config/rails_best_practices.yml
sebastiancaraballo: Well 😅, I tried doing it, but it forces me to use at least one `@`. I think it's because of the multiple contexts, but maybe I'm missing something. Take a look at this flow:
`departments.erb -> _development_metrics/development_metrics_sidebar -> _sidebar.erb`.

app/controllers/concerns/concerns/load_settings.rb
sebastiancaraballo: nice catch

config/rails_best_practices.yml
santib: what forced you to do so? In `departments.erb` it's fine to use it since it's the root view, but then you should pass that as a `local` to the partials https://guides.rubyonrails.org/layouts_and_rendering.html#passing-local-variables

app/controllers/development_metrics_controller.rb
sebastiancaraballo: Actually yes, it's used in several controllers.

config/rails_best_practices.yml
santib: @sebastiancaraballo this comment wasn't resolved

PR #402, opened 2021-01-15
Refactor PART I

sebastiancaraballo:
## What does this PR do?
- This is PART I of a refactor to move the PR size value from the associated `PullRequestSize` model to the `PullRequest` model. By doing this, we avoid unnecessary joins and bugs.
- After deploying it is needed to:
1. Run `rake pull_requests:fill_size` to move size values.
2. Run `rake pull_requests:update_missing_sizes` because there are PRs that do not have an associated size value, but these are available on Github.
3. Run `1` again in order to migrate these newly added values.

* Apart from this, it's worth mentioning that there are PRs from projects that are no longer available, so these will have null size values. So, we have to take into consideration that any query should exclude these null values.

## What's next?
- PART II will be about removing `PullRequestSize` logic and change it to save & update size values at the `PullRequest` table.
- Remove `PullRequestSize` table.

Resolves bugs [#EM-202](https://rootstrap.atlassian.net/browse/EM-202?atlOrigin=eyJpIjoiNDk4OGI1NzJmNGY0NGE4Nzg5NzY2NDhkYWU4ZmViN2EiLCJwIjoiaiJ9) [#EM-203](https://rootstrap.atlassian.net/browse/EM-203?atlOrigin=eyJpIjoiZThkNDU4ZTUwMDcyNGMxOTg4NjliOWY2N2E3ODY5MTAiLCJwIjoiaiJ9)

### DIFF
--- a/app/models/events/pull_request.rb
+++ b/app/models/events/pull_request.rb
@@ -12,6 +12,7 @@
 #  merged_at  :datetime
 #  number     :integer          not null
 #  opened_at  :datetime
+#  size       :integer
 #  state      :enum
 #  title      :text             not null
 #  created_at :datetime         not null
--- a/app/models/external_project.rb
+++ b/app/models/external_project.rb
@@ -4,7 +4,7 @@
 #
 #  id          :bigint           not null, primary key
 #  description :string
-#  enabled     :boolean          default(TRUE)
+#  enabled     :boolean          default(TRUE), not null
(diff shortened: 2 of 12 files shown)


hdamico (approved):
Great job!!

SandroDamilano (approved):
Awesome! 🙌

PR #403, opened 2021-01-18
Refactor PART II

sebastiancaraballo:
## What does this PR do?
- PART II: remove`PullRequestSize` as described at #402.

‼️ Do not merge this into `master` until PART I (#402) tasks are successfully run.

Resolves [#EM-215](https://rootstrap.atlassian.net/browse/EM-215?atlOrigin=eyJpIjoiMThhZTZiNmQ1NWViNDMyZmJjZTM2MzJmYmYwOTIyZDgiLCJwIjoiaiJ9)


santib (approved):
This PR makes sense to me. I think we discussed this in the past but I don't remember what was the reason to have a separate table cc @SandroDamilano @hdamico

db/migrate/20210118145940_drop_pull_request_sizes_table.rb
```diff
@@ -0,0 +1,11 @@
+class DropPullRequestSizesTable < ActiveRecord::Migration[6.0]
+  def change
+    def up
+      drop_table :pull_request_sizes
+    end
+
+    def down
+      raise ActiveRecord::IrreversibleMigration
```
SandroDamilano, line 8: Why do we want to make this migration irreversible? We could easily make it reversible, and I don't think having irreversible migrations is desirable.

SandroDamilano:
@santib There wasn't any strong reason to create a separate table. We just did it that way to make it similar to `MergeTime` and `CompletedReviewTurnaround`, which used their own separate tables.

hdamico (approved):
LGTM!

SandroDamilano (approved):
Great! 👏

PR #414, opened 2021-03-11
Fix active admin initializer halting db at creation for the first time

LALOPSB:
## What does this PR do?

When the dbs are created **for the first time** ActiveAdmin initializers reference them and halts the process because, of course, they don't exist yet, the error is as follows:

```
ActiveAdmin::DatabaseHitDuringLoad: Your file, app/admin/file_ignoring_rules.rb (line 15), 
caused a database error while Active Admin was loading. 
Original error message: FATAL:  database "github_analyzer_development" does not exist
```

To reproduce this error:

```
> rake db:drop
> rake db:create
```

The solution is rescuing the particular exception so when this happens it does not halts the process.


santiagovidal (approved):

config/routes.rb
```diff
@@ -7,7 +7,11 @@
   authenticate :admin_user do
     ExceptionHunter.routes(self)
   end
-  ActiveAdmin.routes(self)
+  begin
```
santib, line 10:
```
ActiveAdmin::DatabaseHitDuringLoad: Your file, app/admin/file_ignoring_rules.rb (line 15), caused a database error while Active Admin was loading. This is most common when your database is missing or doesn't have the latest migrations applied. To prevent this error, move the code to a place where it will only be run when a page is rendered. One solution can be, to wrap the query in a Proc. Original error message: FATAL:  database "github_analyzer_development" does not exist
```

I think the correct solution is using a proc instead of adding this catch logic

matteo95g: Fixed @santib! could you take a look?

santib (approved):
Good job 💪

PR #431, opened 2021-06-14
Replace travis-ci with github actions

matteo95g:
Replace travis-ci with github actions


sebaherrera07 (approved):
Just curious, where did you take the github actions configuration from?

Also, it seems that the checks are failing, is that expected?

matteo95g:
> Just curious, where did you take the github actions configuration from?
> 
> Also, it seems that the checks are failing, is that expected?

I took it from the hotwire playground repo. Yes.. I noticed that I'm trying to fix them!

Gemfile
```diff
@@ -88,6 +88,7 @@ end
 group :test do
   # Adds support for Capybara system testing and selenium driver
   gem 'capybara', '>= 2.15'
+  gem 'fakeredis', require: 'fakeredis/rspec'
```
martinjaimem, line 91: Another option is to actually use Redis in tests. I don't have an opinion about this... LGTM

martinjaimem: You should add the version `0.8.0`

martinjaimem (approved):
Approving 🚀 
I really like where this is going

Before merging, don't forget to set the `fakeredis` version

Gemfile
matteo95g: IMO for testing is ok not to use a real instance of redis

martinjaimem: Yeah, but one could argue, why not doing the same with Postgres?

matteo95g: maybe because of the relationships that exist in the DB, while Redis is only a cache-like memory DB where there are no relationships

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,146 @@
...
+          PGUSER: postgres
+          PGPASSWORD: postgres
+          RAILS_ENV: test
+        run: |
+          bundle exec rspec
+      - name: Report to CodeClimate
+        run: |
+          ./cc-test-reporter after-build --exit-code 0
+        env:
+          CC_TEST_REPORTER_ID: e54f24c60ea55d00f23cba6796704ec633938d432fde7da52a872511496a36b8
```
martinjaimem, line 146: this should be github secret variable

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,146 @@
...
+          sudo apt-get -yqq install libpq-dev
+      - name: Install dependencies
+        run: bundle install
+      - name: Setup CodeClimate
+        run: |
+          curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+          chmod +x ./cc-test-reporter
+          ./cc-test-reporter before-build
+        env:
+          CC_TEST_REPORTER_ID: e54f24c60ea55d00f23cba6796704ec633938d432fde7da52a872511496a36b8
```
martinjaimem, line 123: same here

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,146 @@
...
+          POSTGRES_DB: github_analyzer_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    env:
+      CC_TEST_REPORTER_ID: e54f24c60ea55d00f23cba6796704ec633938d432fde7da52a872511496a36b8
```
martinjaimem, line 74: same

martinjaimem: also, shouldn't this env variable be set just here?

martinjaimem (changes requested):
We should remove private keys from here and use Github secrets

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,142 @@
...
+          chmod +x ./cc-test-reporter
+          ./cc-test-reporter before-build
+      - name: Copy database.yml
+        run: cp config/database.yml.example config/database.yml
+      - name: Setup Database
+        env:
+          PGHOST: localhost
+          PGUSER: postgres
+          PGPASSWORD: postgres
+          RAILS_ENV: test
```
martinjaimem, line 129: we can move these up to line 73

martinjaimem (approved):
LGTM

PR #437, opened 2021-07-01
Fix EDR when several bugs for same environment

enzofab91:
### **Jira cards:**
[EDR-221](https://rootstrap.atlassian.net/browse/EM-221) - EDR not taking updated bugs into account
[EDR-222](https://rootstrap.atlassian.net/browse/EM-222) - EDR showing empty string when environment is null

### **Description:**
This PR fixes a bug that is getting wrong quantity issues when are several bugs in the same period.
It also fixes the view when environment is `null`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -149,7 +149,7 @@ GEM
     faraday (1.1.0)
       multipart-post (>= 1.2, < 3)
       ruby2_keywords
-    ffi (1.11.3)
+    ffi (1.15.3)
     formtastic (3.1.5)
       actionpack (>= 3.2.13)
     formtastic_i18n (0.6.0)
@@ -349,7 +349,7 @@ GEM
     safe_yaml (1.0.5)
     sass-rails (6.0.0)
       sassc-rails (~> 2.1, >= 2.1.1)
-    sassc (2.2.1)
+    sassc (2.4.0)
       ffi (~> 1.9)
     sassc-rails (2.1.2)
       railties (>= 4.0.0)
--- a/app/services/metrics/defect_escape_rate/per_project.rb
+++ b/app/services/metrics/defect_escape_rate/per_project.rb
@@ -28,8 +28,8 @@ module Metrics
       end
 
       def bugs_by_environment(bug_issues)
-        bug_issues.each_with_object({}) do |bug, result|
-          result[bug.environment] = +1
+        bug_issues.each_with_object(Hash.new(0)) do |bug, result|
+          result[bug.environment] += 1
         end
       end
     end
--- a/app/views/development_metrics/defect_escape_rate/_values_details.erb
+++ b/app/views/development_metrics/defect_escape_rate/_values_details.erb
@@ -16,10 +16,9 @@
       <hr>
       <div>
         <h5>Bugs by environment:</h5>
-        <% metric[:values].each do |metric| %>
+        <% metric[:values].each do |environment, bugs| %>
           <div class="item-list">
-            <%= metric.first.to_s.capitalize %>:
-              <%= metric.second.to_s %>
+            <%= environment.to_s.capitalize.presence || 'None' %>: <%= bugs %>
           </div>
           <div></div>
         <% end %>
--- a/spec/controllers/development_metrics_controller_spec.rb
+++ b/spec/controllers/development_metrics_controller_spec.rb
@@ -1,11 +1,12 @@
 require 'rails_helper'
 
-RSpec.describe DevelopmentMetricsController, type: :controller do
+describe DevelopmentMetricsController, type: :controller do
   fixtures :departments, :languages
 
   let(:ruby_lang) { Language.find_by(name: 'ruby') }
   let(:project) { create(:project, name: 'rs-metrics', language: ruby_lang) }
   let!(:jira_project) { create(:jira_project, project: project) }
+  let(:beginning_of_day) { Time.zone.today.beginning_of_day }
 
   describe '#index' do
     context 'when metric params are empty' do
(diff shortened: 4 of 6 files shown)


app/services/metrics/defect_escape_rate/per_project.rb
```diff
@@ -27,9 +27,14 @@ def query(interval)
         ]]
       end
 
+      # :reek:DuplicateMethodCall
       def bugs_by_environment(bug_issues)
         bug_issues.each_with_object({}) do |bug, result|
```
horacio, line 32: Maybe we can set a default for these keys? Something like...

```suggestion
        bug_issues.each_with_object(Hash.new(1)) do |bug, result|
```

Gemfile.lock
```diff
@@ -149,7 +149,7 @@ GEM
     faraday (1.1.0)
       multipart-post (>= 1.2, < 3)
       ruby2_keywords
-    ffi (1.11.3)
+    ffi (1.15.3)
```
santib, line 152: in general, we shouldn't upgrade gems as part of other tickets

app/services/metrics/defect_escape_rate/per_project.rb
```diff
@@ -27,9 +27,14 @@ def query(interval)
         ]]
       end
 
+      # :reek:DuplicateMethodCall
```
santib, line 30: I'd prefer to have these comments in `.reek.yml` rather than polluting the actual code with them

app/services/metrics/defect_escape_rate/per_project.rb
santib: Agree with Hora, but actually, I think it should start in 0 by default and always add 1

```ruby
bug_issues.each_with_object(Hash.new(0)) do |bug, result|
  result[bug.environment] += 1
end
```

app/views/development_metrics/defect_escape_rate/_values_details.erb
```diff
@@ -18,7 +18,11 @@
         <h5>Bugs by environment:</h5>
         <% metric[:values].each do |metric| %>
           <div class="item-list">
+            <% if metric.first -%>
             <%= metric.first.to_s.capitalize %>:
+            <% else -%>
+              None:
+            <% end -%>
```
santib, line 25:
```suggestion
            <%= metric.first.to_s.capitalize.presence || 'None:' %>:
```

Apart from that, I don't remember if we have presenters in this project, if we do have it might go there.

app/views/development_metrics/defect_escape_rate/_values_details.erb
```diff
@@ -18,7 +18,11 @@
         <h5>Bugs by environment:</h5>
         <% metric[:values].each do |metric| %>
           <div class="item-list">
+            <% if metric.first -%>
             <%= metric.first.to_s.capitalize %>:
+            <% else -%>
+              None:
```
santib, line 24: we are missing a test checking for `None:`

app/views/development_metrics/defect_escape_rate/_values_details.erb
```diff
@@ -18,7 +18,11 @@
         <h5>Bugs by environment:</h5>
         <% metric[:values].each do |metric| %>
```
santiagovidal, line 19: Instead of doing `metric.first` and `metric.second`, we can define this loop like:

`metric[:values].each do |environment, bugs|`

It's easier to follow

app/services/metrics/defect_escape_rate/per_project.rb
enzofab91: Nice!

app/views/development_metrics/defect_escape_rate/_values_details.erb
enzofab91: We don't have presenters in this project yet.

Gemfile.lock
enzofab91: When I did `bundle install` I have to update this gem in order to work. You think I have to remove this file from PR?

app/views/development_metrics/defect_escape_rate/_values_details.erb
enzofab91: Nice, done!

app/views/development_metrics/defect_escape_rate/_values_details.erb
enzofab91: Done!

.rspec
```diff
@@ -1 +1 @@
---require spec_helper
+--require rails_helper
```
santib, line 1: I think this was fine, we should require `spec_helper` here. Why was the change?

Gemfile.lock
santib: we can leave it this time

app/views/development_metrics/defect_escape_rate/_values_details.erb
```diff
@@ -16,10 +16,10 @@
       <hr>
       <div>
         <h5>Bugs by environment:</h5>
-        <% metric[:values].each do |metric| %>
+        <% metric[:values].each do |environment, bugs| %>
           <div class="item-list">
-            <%= metric.first.to_s.capitalize %>:
-              <%= metric.second.to_s %>
+            <%= environment.to_s.capitalize.presence || 'None' %>:
+            <%= bugs.second.to_s %>
```
santib, line 22: shouldn't this be
```suggestion
            <%= bugs %>
```

now that you changed `|metric|` for `|environment, bugs|` ?

spec/controllers/development_metrics_controller_spec.rb
```diff
@@ -1,11 +1,12 @@
-require 'rails_helper'
```
santib, line 1: we should keep this line here

.rspec
enzofab91: `rails_helper` includes `spec_helper`, and allows me to remove `require 'rails_helper'` from `spec` files.

spec/controllers/development_metrics_controller_spec.rb
enzofab91: Related to the comment above.

spec/controllers/development_metrics_controller_spec.rb
```diff
@@ -44,10 +45,82 @@
...
+          before do
+            subject
+          end
+
+          it 'has a successful response' do
+            expect(response).to be_successful
+          end
+
+          it 'render EDR metric with correct production issues' do
+            expect(response.body).to include('Production:' + NEWLINE_WHITESPACES +
```
santib, line 107: the `NEWLINE_WHITESPACES` thing looks fragile. Why not doing `/Production:\s*#{production_jira_bugs.count}/` ?

Also don't use `+` for string interpolation, instead use `"#{} #{}`

app/views/development_metrics/defect_escape_rate/_values_details.erb
enzofab91: Haha yes 😅

santib, line 22:
```suggestion
            <%= environment.to_s.capitalize.presence || 'None' %>: <%= bugs.second.to_s %>
```

is there a reason to not have it like that?

.rspec
santib: https://github.com/rspec/rspec-rails/issues/1180

app/views/development_metrics/defect_escape_rate/_values_details.erb
enzofab91: No reason why, I'll change it.

santib (approved):
good job 💪

santiagovidal (approved):

horacio (approved):

PR #441, opened 2021-07-08
Add Product model

enzofab91:
## What does this PR do?

Creates a new entity `Product`, which represents a product from a client. This product is managed in a Jira project and has some KPIs. At the moment, only Defect Escape Rate is supported. In addition, a product have several Github repositories (`project` entity).

This PR is one part of the full feature. Next steps includes adding this information in the frontend.

Resolves [EM-223](https://rootstrap.atlassian.net/browse/EM-223) (in part).

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -1,0 +1,23 @@
+# == Schema Information
+#
+# Table name: products
+#
+#  id          :bigint           not null, primary key
+#  description :string
+#  name        :string           not null
+#
+# Indexes
+#
+#  index_products_on_name  (name)
+#
+
+class Product < ApplicationRecord
+  has_one :jira_project, dependent: :destroy
+
+  has_many :projects, dependent: :destroy
+  has_many :metrics, as: :ownable, dependent: :destroy
+
+  validates :name, presence: true, uniqueness: true
+
+  accepts_nested_attributes_for :jira_project
+end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -11,10 +11,16 @@
 #  updated_at  :datetime         not null
 #  github_id   :integer          not null
 #  language_id :bigint
+#  product_id  :bigint
 #
 # Indexes
 #
 #  index_projects_on_language_id  (language_id)
+#  index_projects_on_product_id   (product_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (product_id => products.id)
 #
 
 class Project < ApplicationRecord
@@ -26,6 +32,7 @@ class Project < ApplicationRecord
   }
 
   belongs_to :language
+  belongs_to :product, optional: true
 
   has_many :events, dependent: :destroy
   has_many :pull_requests,
@@ -49,13 +56,13 @@ class Project < ApplicationRecord
            source: :user
 
   has_one :code_climate_project_metric, dependent: :destroy
-  has_one :jira_project, dependent: :destroy
-  accepts_nested_attributes_for :jira_project
 
   validates :github_id, presence: true, uniqueness: true
   validates :relevance, inclusion: { in: relevances.keys }
 
   before_validation :set_default_language, on: :create
+
+  delegate :jira_project, to: :product
 
   scope :open_source, lambda {
     internal.with_language.where(is_private: false)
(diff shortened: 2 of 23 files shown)


app/models/product.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class Product < ApplicationRecord
+  belongs_to :jira_project
+
+  has_many :projects, dependent: :destroy
+  has_many :metrics, as: :ownable, dependent: :destroy
+
+  validates :name, presence: true, uniqueness: true
+
+  # accepts_nested_attributes_for :jira_project
```
sebastiancaraballo, line 28: 👀

app/models/project.rb
```diff
@@ -115,7 +120,11 @@ def organization_name
   end
 
   def jira_key
-    jira_project&.jira_project_key
+    product&.jira_project&.jira_project_key
```
sebastiancaraballo, line 123: What about adding a delegate?

app/models/project.rb
```diff
@@ -115,7 +120,11 @@ def organization_name
...
+  end
+
+  def jira_project?
```
sebastiancaraballo, line 126: Is this being used?

enzofab91: Yes, when creating developer metrics chartkick. If the project has a jira project it adds the metrics related.

app/models/project.rb
enzofab91: Done!

spec/models/product_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+#
+#  fk_rails_...  (jira_project_id => jira_projects.id)
+#
+
+require 'rails_helper'
+
+describe Product, type: :model do
+  subject { build :product }
+
+  context 'validations' do
```
sebastiancaraballo, line 25:
```
describe 'validations'
  context 'with valid attrs...'
```

sebastiancaraballo (approved):

app/models/product.rb
```diff
@@ -0,0 +1,29 @@
...
+#  index_products_on_jira_project_id  (jira_project_id)
+#  index_products_on_name             (name)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (jira_project_id => jira_projects.id)
+#
+
+class Product < ApplicationRecord
+  belongs_to :jira_project
```
santib, line 21: hmm so the product BELONGS to a JIRA project? Doesn't sound correct to me. Instead, I think the `Product` to be the root of all models, and that the JIRA project BELONGS to a product.. thoughts?

app/models/project.rb
```diff
@@ -26,6 +32,7 @@ class Project < ApplicationRecord
   }
 
   belongs_to :language
+  belongs_to :product, optional: true
```
santib, line 35: why is this optional?

app/services/builders/chartkick/development_metrics.rb
```diff
@@ -51,7 +51,7 @@ def defect_escape_rate_entities
         end
 
         def project_has_jira_board_associated?(project_id)
-          ::Project.find(project_id).jira_project.present?
+          ::Project.find(project_id).jira_project?
```
santib, line 54: I prefer to have `.present?` here instead of creating a method in the model

app/models/product.rb
enzofab91: I didn't think of model hierarchy, just created the relationship.  I'll change it so we could have a product even if the jira project haven't been created yet.

app/services/builders/chartkick/development_metrics.rb
enzofab91: Done!

app/models/project.rb
enzofab91: This is because when we receive the webhooks from GH, if the project don't exist it is created. The payload does not have any information about the product and it fails to create it. Maybe can be change it to only find the `project` and the admin has to create the project manually before start reporting the events. WDYT?

matteo95g (approved):

app/models/project.rb
santib: @enzofab91 for the future, I don't think it's fine to merge PRs with open discussions

PR #446, opened 2021-07-22
Add tooltip to metrics

enzofab91:
## What does this PR do?
It creates a new entity `metric_definition` with the `name` of a metric and an `explanation` of what it is. This information is handled in the admin panel and can be seen in the frontend.

Each metric has a question mark that when it is hovered it shows the explanation, as can be seen in next image:

Screen Shot 2021-07-22 at 14 19 20

This image shows the admin panel for metric definition:

Screen Shot 2021-07-22 at 14 20 36

Resolves [EM-89](https://rootstrap.atlassian.net/browse/EM-89): Add tooltips to explain each metric

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -70,3 +70,16 @@ canvas[style] {
     }
   }
 }
+
+.metric_tooltip {
+  font-size: x-large;
+  color: white;
+  background:black;
+  border-radius:50%;
+  height: 26px;
+  width: 26px;
+  line-height: 26px;
+  display: inline-block;
+  text-align: center;
+  margin-right: 6px;
+}
--- a/app/javascript/packs/toggleDetailsMetric.js
+++ b/app/javascript/packs/toggleDetailsMetric.js
@@ -20,3 +20,7 @@ $(document).on('turbolinks:load', function() {
     $('.details-button').show();
   }
 });
+
+$(function () {
+  $('[data-toggle="tooltip"]').tooltip();
+});
--- a/app/models/metric_definition.rb
+++ b/app/models/metric_definition.rb
@@ -1,0 +1,25 @@
+# == Schema Information
+#
+# Table name: metric_definitions
+#
+#  id          :bigint           not null, primary key
+#  code        :enum             not null
+#  explanation :string           not null
+#  name        :string           not null
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+
+class MetricDefinition < ApplicationRecord
+  enum code: { review_turnaround: 'review_turnaround',
+               blog_visits: 'blog_visits',
+               merge_time: 'merge_time',
+               blog_post_count: 'blog_post_count',
+               open_source_visits: 'open_source_visits',
+               defect_escape_rate: 'defect_escape_rate',
+               pull_request_size: 'pull_request_size' }
+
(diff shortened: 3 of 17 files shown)


app/models/metric_definition.rb
```diff
@@ -0,0 +1,25 @@
...
+class MetricDefinition < ApplicationRecord
+  enum code: { review_turnaround: 'review_turnaround',
+               blog_visits: 'blog_visits',
+               merge_time: 'merge_time',
+               blog_post_count: 'blog_post_count',
+               open_source_visits: 'open_source_visits',
+               defect_escape_rate: 'defect_escape_rate',
+               pull_request_size: 'pull_request_size' }
+
+  validates :code, inclusion: { in: codes.keys }, uniqueness: true
```
matteo95g, line 22: I think the inclusion is not necessary because it's an enum

app/views/development_metrics/projects.erb
```diff
@@ -32,7 +35,10 @@
     </div>
 
     <div class="metrics shadow-box">
-      <h4 class="p-3">Time to merge</h4>
+      <div class='row'>
+        <h4 class="p-3" style='margin-left:10px;'><%= @merge_time_definition && @merge_time_definition[:name] %></h4>
```
matteo95g, line 39: you can do only `@merge_time_definition[:name]` because if the value don't exist it will return nil

matteo95g: also, don't you need to pass the name of the metric in `:name`? how do you know what metric will show up there?

app/models/metric_definition.rb
enzofab91: You're right! I'll remove it.

app/javascript/packs/toggleDetailsMetric.js
```diff
@@ -20,3 +20,7 @@ $(document).on('turbolinks:load', function() {
     $('.details-button').show();
   }
 });
+
+$(function () {
+  $('[data-toggle="tooltip"]').tooltip();
+});
```
lauperalti, line 26:
```suggestion
});

```

app/assets/stylesheets/application.scss
```diff
@@ -69,4 +69,17 @@ canvas[style] {
...
+  color: white;
+  background:black;
+  border-radius:50%;
+  height: 26px;
+  width: 26px;
+  line-height: 26px;
+  display: inline-block;
+  text-align: center;
+  margin-right: 6px;
+}
```
lauperalti, line 85:
```suggestion
}

```

app/models/metric_definition.rb
```diff
@@ -0,0 +1,25 @@
...
+class MetricDefinition < ApplicationRecord
+  enum code: { review_turnaround: 'review_turnaround',
+               blog_visits: 'blog_visits',
+               merge_time: 'merge_time',
+               blog_post_count: 'blog_post_count',
+               open_source_visits: 'open_source_visits',
+               defect_escape_rate: 'defect_escape_rate',
+               pull_request_size: 'pull_request_size' }
+
+  validates :code, uniqueness: true
```
lauperalti, line 22: is the _code_ non-null? can the validation `presence: true` be necessary?

lauperalti (approved):

matteo95g (approved):

PR #450, opened 2021-07-30
Add feature pull request comments - 2 of 2

lauperalti:
## What does this PR do?
This PR is the second part of the full feature.

Resolves #[EM-219](https://rootstrap.atlassian.net/browse/EM-219) (Second Step)
Handle issue_comment event for the PR comments written by reviewers(assigned to PR) and use it to calculate time to 2nd review.

### DIFF
--- a/app/models/events/pull_request_comment.rb
+++ b/app/models/events/pull_request_comment.rb
@@ -42,5 +42,13 @@ module Events
 
     validates :state, inclusion: { in: states.keys }
     validates :github_id, :body, :opened_at, presence: true
+
+    after_create :build_review_or_comment_turnaround
+
+    private
+
+    def build_review_or_comment_turnaround
+      Builders::ReviewOrCommentTurnaround.call(review_request, pull_request, self)
+    end
   end
 end
--- a/app/models/events/pull_request_comment.rb
+++ b/app/models/events/pull_request_comment.rb
@@ -1,0 +1,46 @@
+# == Schema Information
+#
+# Table name: pull_request_comments
+#
+#  id                :bigint           not null, primary key
+#  body              :string
+#  deleted_at        :datetime
+#  opened_at         :datetime         not null
+#  state             :enum             default("created")
+#  created_at        :datetime         not null
+#  updated_at        :datetime         not null
+#  github_id         :integer
+#  owner_id          :bigint
+#  pull_request_id   :bigint           not null
+#  review_request_id :bigint
+#
+# Indexes
+#
+#  index_pull_request_comments_on_owner_id           (owner_id)
+#  index_pull_request_comments_on_pull_request_id    (pull_request_id)
+#  index_pull_request_comments_on_review_request_id  (review_request_id)
+#  index_pull_request_comments_on_state              (state)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (owner_id => users.id)
+#  fk_rails_...  (pull_request_id => pull_requests.id)
+#
+
+module Events
+  class PullRequestComment < ApplicationRecord
+    acts_as_paranoid
+
+    enum state: { created: 'created', edited: 'edited', deleted: 'deleted' }
+
+    has_many :events, as: :handleable, dependent: :destroy
+    belongs_to :owner, class_name: 'User',
+                       foreign_key: :owner_id,
+                       inverse_of: :owned_pull_request_comments
+    belongs_to :pull_request, class_name: 'Events::PullRequest'
+    belongs_to :review_request
+
+    validates :state, inclusion: { in: states.keys }
+    validates :github_id, :body, :opened_at, presence: true
+  end
+end
--- a/app/models/events/pull_request_comment.rb
+++ b/app/models/events/pull_request_comment.rb
@@ -16,6 +16,7 @@
 #
 # Indexes
 #
+#  index_pull_request_comments_on_deleted_at         (deleted_at)
 #  index_pull_request_comments_on_owner_id           (owner_id)
 #  index_pull_request_comments_on_pull_request_id    (pull_request_id)
 #  index_pull_request_comments_on_review_request_id  (review_request_id)
--- a/app/services/builders/review_or_comment_turnaround.rb
+++ b/app/services/builders/review_or_comment_turnaround.rb
@@ -1,0 +1,46 @@
+module Builders
+  class ReviewOrCommentTurnaround < BaseService
+    def initialize(review_request, pull_request, action)
+      @review_request = review_request
+      @pull_request = pull_request
+      @action = action
+    end
+
+    def call
+      build_review_turnaround
+      build_completed_review_turnaround
+    end
+
+    private
+
+    def build_review_turnaround
+      return unless uniq_review_on_review_request?
+
+      Builders::ReviewTurnaround.call(@review_request)
+    end
+
+    def build_completed_review_turnaround
+      return unless second_review_with_different_users? && uniq_review_on_review_request?
+
+      Builders::CompletedReviewTurnaround.call(@action)
+    end
+
+    def second_review_with_different_users?
+      reviews_or_comments_different_users == 2
+    end
+
+    def reviews_or_comments_different_users
+      count_reviews = @pull_request .reviews.distinct.pluck(:owner_id)
+      count_comments = @pull_request .pull_request_comments.distinct.pluck(:owner_id)
+      (count_reviews | count_comments).count
+    end
+
+    def uniq_review_on_review_request?
+      reviews_or_comments_on_review_request.equal?(1)
+    end
+
+    def reviews_or_comments_on_review_request
+      @review_request.reviews.count + @review_request.pull_request_comments.count
+    end
+  end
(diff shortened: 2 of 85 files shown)


spec/services/builders/review_or_comment_turnaround_spec.rb
```diff
@@ -0,0 +1,149 @@
+require 'rails_helper'
+
+RSpec.describe Builders::ReviewOrCommentTurnaround do
```
enzofab91, line 3:
```suggestion
describe Builders::ReviewOrCommentTurnaround do
```

spec/services/builders/review_or_comment_turnaround_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+    shared_examples 'when a different user make the second action' do
+      it 'builds the complete review turnaround' do
+        expect { second_action }.to change { CompletedReviewTurnaround.count }.by(1)
+      end
+    end
+
+    shared_examples 'when the owner of the second action makes a third action' do
+      before { second_action }
+
+      it 'does not build the completd review turnaround' do
```
enzofab91, line 117:
```suggestion
      it 'does not build the completed review turnaround' do
```

enzofab91 (approved):
Great work! 🚀 

Left some minor comments.

app/models/events/pull_request_comment.rb
```diff
@@ -43,5 +43,13 @@ class PullRequestComment < ApplicationRecord
 
     validates :state, inclusion: { in: states.keys }
     validates :github_id, :body, :opened_at, presence: true
+
+    after_create :build_review_or_comment_turnaround
```
santib, line 47: I'm not a fan of model callbacks for multiple reasons, there are many blog posts written about their cons, here is one example https://engineering.gusto.com/the-rails-callbacks-best-practices-used-at-gusto/

>Conclusion
With great power comes great responsibility, and Rails callbacks are no exception. They are expressive tools but also come with drawbacks.
>
>We are generally moving away from using callbacks and toward using rich service objects in our applications, but we still have many callbacks that exist. We still have a lot left to do, but we’ve found that the code that follows these guidelines requires less maintenance.

Here are a couple more links:
- https://blog.arkency.com/2016/05/domain-events-over-active-record-callbacks/ 
- https://twitter.com/_solnic_/status/974214728066260993

In a few words, we are polluting a domain object with complex creational logic (this is a bit advanced but I highly recommend it https://www.infoq.com/minibooks/domain-driven-design-quickly/)

Wouldn't it be possible to have this logic in another place (e.g. a service where the PullRequest is created)

app/services/builders/review_or_comment_turnaround.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def reviews_or_comments_different_users
+      count_reviews = @pull_request .reviews.distinct.pluck(:owner_id)
+      count_comments = @pull_request .pull_request_comments.distinct.pluck(:owner_id)
+      (count_reviews | count_comments).count
+    end
+
+    def uniq_review_on_review_request?
+      reviews_or_comments_on_review_request.equal?(1)
```
santib, line 39: maybe this should be memoized since it's called more than once?
```suggestion
      @uniq_review_on_review_request ||= reviews_or_comments_on_review_request.equal?(1)
```

app/services/builders/review_or_comment_turnaround.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      Builders::CompletedReviewTurnaround.call(@action)
+    end
+
+    def second_review_with_different_users?
+      reviews_or_comments_different_users == 2
+    end
+
+    def reviews_or_comments_different_users
+      count_reviews = @pull_request .reviews.distinct.pluck(:owner_id)
```
santib, line 33: in this line and the line below there is an extra space before the `.reviews`

app/services/github_service.rb
```diff
@@ -29,13 +29,19 @@ def handle_event
...
+    return event.gsub('issue_', 'pull_request_') if pull_request_comment?
+
     return event.gsub('pull_request_', '') unless event == 'pull_request'
 
     event
   end
 
   def handleable_event?
-    Event::TYPES.include?(@event)
+    Event::TYPES.include?(@event) || pull_request_comment?
```
santib, line 40: Is there any reason to not have `pull_request_comment` included in `Event::TYPES` ?

db/structure.sql
```diff
@@ -631,9 +631,9 @@ CREATE TABLE public.external_pull_requests (
     external_project_id bigint NOT NULL,
     created_at timestamp(6) without time zone NOT NULL,
     updated_at timestamp(6) without time zone NOT NULL,
+    number integer,
     opened_at timestamp without time zone,
-    state public.external_pull_request_state,
-    number integer
+    state public.external_pull_request_state
```
santib, line 636: what are these changes? I don't see any migration added in this PR

spec/services/builders/completed_review_turnaround_spec.rb
```diff
@@ -10,20 +10,38 @@
...
+        expect(described_class.call(review).value).to eq(correct_value)
+      end
+    end
+
+    context 'when second review is a review' do
+      let!(:review) do
+        create(:review, owner: santib, project: project, pull_request: pr, review_request: rr)
+      end
+
+      it_behaves_like 'returns'
```
santib, line 33: this doesn't look very descriptive 😅

app/models/events/pull_request_comment.rb
santib: given that we are already using callbacks in the project, I'm good leave this as it is, but keep it in mind for future PRs that could be a good idea to move away from model callbacks

spec/services/github_service_spec.rb
```diff
@@ -281,5 +281,82 @@
...
+          expect {
+            subject
+          }.to change { pull_request_comment.reload.state }
+            .from(pull_request_comment.state).to('edited')
+        end
+      end
+
+      context 'when the action is deleted' do
+        it 'sets deleted' do
+          change_action_to('deleted')
```
santib, line 346: these kind of code that is actually setting up the test conditions would look better in `let` or `before` blocks in this case

spec/services/builders/review_turnaround_spec.rb
```diff
@@ -117,8 +183,19 @@
                pull_request: pr,
                created_at: wednesday_past_week)
       end
+    end
 
-      let!(:review) do
+    shared_examples 'for thursday and two weeks later on wednesday' do
```
santib, line 188: I think the shared examples name should be something so that when I read
`it behave like x` I understand what it's testing, here you are describing like if it were a `context`

santib (approved):
Good job, left some comments. Let me know when you have responded/fixed things and I'll take another look. Also be careful on abusing of `shared_examples` test code is not the same as app code, trying to be DRY is sometimes dangerous. More specifically I value test readability over DRYness. Some links below:
- https://thevaluable.dev/dry-principle-cost-benefit-example/
- https://twitter.com/avdi/status/1351985294711599104 (esto es en los tests en realidad)
- https://anti-pattern.com/water-your-controllers-they-are-too-dry
- https://blog.ragnarson.com/2015/09/30/problems-with-typical-rails-controllers-and-before-actions.html

app/models/events/pull_request_comment.rb
lauperalti: Oh! Interesting information, Thank you very much. I will of course keep it in mind for future developments.

app/services/github_service.rb
lauperalti: Actually, I decided to go with this implementation because the event itself has the name "issue_comment", and it is used not only for PR comments but for Github Issues comments. And we don't want to include that ones but the ones that have "pull_request" in their json.

db/structure.sql
lauperalti: IDK, seems like it was automatically generated, may be because some local merge I did. Looks likejust changed the order of the lines.

lauperalti:
@santib Thank you very much for all your comments and suggestions. I have just pushed some of the corrections you suggest and answered some of your questions. 
Also, I have taken a look at the articles you mention and will of course keep them in mind for future developments.

spec/models/events/review_spec.rb
```diff
@@ -54,81 +54,15 @@
...
-    describe '#build_review_turnaround' do
+    describe '#build_review_or_comment_turnaround' do
       context 'when a review is created' do
         before { travel_to(Time.zone.today.beginning_of_day) }
         let(:review_request) { create(:review_request) }
         let(:review) { create :review, review_request: review_request }
 
-        it 'creates a review turnaround with the correct values' do
+        it 'calls the review/comment builder helper' do
+          expect(Builders::ReviewOrCommentTurnaround). to receive(:call)
```
santib, line 64:
```suggestion
          expect(Builders::ReviewOrCommentTurnaround).to receive(:call)
```

santib (approved):
Great job 👏

PR #451, opened 2021-08-03
Development cycle metric - Part 2 of 2

enzofab91:
## What does this PR do?

This PR is the second part of the full feature. In this PR the metric is added to the frontend.
Also, a small fix to tooltip metrics were found.

Now in the product view a summary of both metrics are shown (last four weeks), as can be seen in next image:

Screen Shot 2021-08-03 at 19 59 08

When user click in 'See details' button, a chart with more details is shown and the period is variable, as can be seen in next image:

Screen Shot 2021-08-03 at 19 40 48

Resolves [EM-226](https://rootstrap.atlassian.net/browse/EM-226): See project metric: Development Cycle.

### DIFF
--- a/app/controllers/development_metrics_controller.rb
+++ b/app/controllers/development_metrics_controller.rb
@@ -2,6 +2,8 @@ class DevelopmentMetricsController < ApplicationController
   layout 'sidebar_metrics'
   include LoadSettings
 
+  PRODUCTS_ACTION = 'products'.freeze
+
   def index; end
 
   def products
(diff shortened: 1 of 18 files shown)


app/controllers/development_metrics_controller.rb
```diff
@@ -54,10 +56,19 @@ def build_metrics(entity_id, entity_name)
   end
 
   def build_product_metrics(entity_id, entity_name)
+    @has_jira_project = Product.find(product.id).jira_project&.present?
```
santiagovidal, line 59: Why search for the product if you already have it in the `product` variable?

app/controllers/development_metrics_controller.rb
```diff
@@ -54,10 +56,19 @@ def build_metrics(entity_id, entity_name)
...
+
     metrics = Builders::Chartkick::DevelopmentMetrics.const_get(entity_name)
                                                      .call(entity_id, metric_params[:period])
-    @defect_escape_rate = metrics[:defect_escape_rate]
-    @metric_definition = MetricDefinition.find_by(code: :defect_escape_rate)
+
+    defect_escape_rate(metrics)
+    development_cycle(metrics)
+
+    @query_metric_name = metric_params[:name]
```
santiagovidal, line 71: You can set this variable at the beginning of the method and use it in the `show_metrics?` method

app/controllers/development_metrics_controller.rb
```diff
@@ -66,6 +77,38 @@ def build_metrics_definitions
     @pull_request_size_definition = MetricDefinition.find_by(code: :pull_request_size)
   end
 
+  def show_metrics?
+    metrics_to_show if metric_params[:name].present?
+
+    product_action = action == PRODUCTS_ACTION
+
+    @show_defect_escape_rate = product_action || @show_defect_escape_rate ? true : false
```
santiagovidal, line 85: This line looks confusing, I think we can do `@show_defect_escape_rate ||= product_action`

matteo95g (approved):

app/controllers/development_metrics_controller.rb
```diff
@@ -66,6 +77,48 @@ def build_metrics_definitions
...
+    when :development_cycle
+      @show_development_cycle = true
+    end
+  end
+
+  def defect_escape_rate(metrics)
+    return unless @show_defect_escape_rate
+
+    @defect_escape_rate = metrics[:defect_escape_rate]
+    @metric_definition = MetricDefinition.find_by(code: :defect_escape_rate)
```
lauperalti, line 100: ` @metric_definition` this name does not seem very descriptive since, as far as I understand, it refers only to defect escape rate definition. For example you used `@development_cycle_definition` below, that is more accurate. 

if it involves a lot of change don't worry. Just personal taste here.

enzofab91: Done! I haven't notice but is better for readability.

app/services/metrics/development_cycle/per_product.rb
```diff
@@ -0,0 +1,42 @@
...
+          query(interval).map do |product, metric_value|
+            Metric.new(product, interval.first, metric_value)
+          end
+        end
+      end
+
+      def query(interval)
+        product = Product.find(@entity_id)
+
+        resolved_issues = product.jira_project.jira_issues.where(resolved_at: interval)
```
lauperalti, line 18: I found that this method is a bit long and also `resolved_issues` is used too many times in and outside it.

If more code is needed in the future, before code smells, may be is good to consider for example doing a new **def** and use the `@resolved_issues `.

WDYT?

lauperalti (approved):
LGTM 👏 🎉

PR #452, opened 2021-08-09
Remove Jira project entity

enzofab91:
## What does this PR do?

This PR removes `JiraProject` entity, by adding the `jira_project_key` directly to `product` entity.

Resolves 
[EM-229](https://rootstrap.atlassian.net/browse/EM-229): 500 error when creating JIRA project.
[EM-230](https://rootstrap.atlassian.net/browse/EM-230): Add JIRA Key field to Product, remove JIRA Project entity.


app/models/jira_project.rb
```diff
@@ -1,30 +0,0 @@
...
-# Indexes
-#
-#  index_jira_projects_on_product_id  (product_id)
-#
-# Foreign Keys
-#
-#  fk_rails_...  (product_id => products.id)
-#
-
-class JiraProject < ApplicationRecord
```
santib, line 22: Why are we removing the `JiraProject`? I think it might make sense to keep it (maybe rename it to `JiraBoard`), but  having `JiraIssue` relate to a `Product` doesn't look that good to me. In the future we might even have N JiraBoards for 1 Product...

enzofab91:
We decided to keep `JiraProject` entity and renamed to `JiraBoard`.

PR #454, opened 2021-08-10
Rename Jira Project to Jira Board

enzofab91:
## What does this PR do?

This PR renames `JiraProject` entity to `JiraBoard`. Also, when creating a Product in the admin tab it creates internally the `JiraBoard` if a Jira key is added.

Resolves 
[EM-229](https://rootstrap.atlassian.net/browse/EM-229): 500 error when creating JIRA project.
[EM-230](https://rootstrap.atlassian.net/browse/EM-230): Add JIRA Key field to Product, remove JIRA Project entity.


app/services/metrics/defect_escape_rate/per_project.rb
```diff
@@ -1,37 +0,0 @@
-module Metrics
```
santib, line 1: why was it deleted? how is it related to the rename of JiraBoard?

santib (approved):
approved, left just one question to understand better the scope of this work

app/services/metrics/defect_escape_rate/per_project.rb
enzofab91: Is not related to this rename. I have noticed that this file is unused because DER metric is now by product non by project.

santib: ok for the future, especially for real projects, consider creating separate PRs (and even creating separate tickets) when tackling unrelated stuff

lauperalti (approved):
👏 💪

PR #455, opened 2021-08-11
Remove unused sidekiq job

enzofab91:
## What does this PR do?
This PR removes an unused old job call in the `sidekiq` configuration file.

Resolves [EM-231](https://rootstrap.atlassian.net/browse/EM-231): NameError: uninitialized constant WeeklyMetricsJob


config/sidekiq.yml
```diff
@@ -4,4 +4,3 @@ production:
   :concurrency: 10
 :queues:
   - default
-  - weekly_metrics
```
sebastiancaraballo, line 7: I assume you tested this. Why does deleting the queue solve the issue?

sebastiancaraballo (approved):

config/sidekiq.yml
enzofab91: It won't haha, I had remove it manually from sidekiq cron in order to stop running. I think it is also correct to remove it from this file just in case in the future the app is restarted and cron reads this config, isn't it?

santiagovidal (approved):

PR #461, opened 2021-08-13
Development cycle empty state

enzofab91:
## What does this PR do?
This PR adds an empty state when no issues are reported in the selected period for development cycle metric with the same behaviour as defect escape rate, as can be seen in next image:

Screen Shot 2021-08-13 at 11 53 08

Resolves [EM-237](https://rootstrap.atlassian.net/browse/EM-237): Development cycle empty state


lauperalti (approved):
Good idea!

santiagovidal (approved):

PR #466, opened 2021-08-23
Adds retry when github id is already been taken

enzofab91:
## What does this PR do?
In [this](https://github.com/rootstrap/rs-code-review-metrics/pull/464) previous PR, talking with @santib we figured out that `find_or_create_by` function has a strange behaviour when the model has a uniqueness validation (more [info](https://github.com/rails/rails/issues/36027)). So that we decided to leave it as it is and retry the job in this case with a custom exception.

Resolves [EM-239](https://rootstrap.atlassian.net/browse/EM-239): ActiveRecord::RecordInvalid: Validation failed: Github has already been taken


santib (approved):
good job 💯

PR #473, opened 2021-09-02
Refactor Project to Repository - Part 2

enzofab91:
## What does this PR do?
This PR is the second part of the full feature. The first part is [here](https://github.com/rootstrap/rs-code-review-metrics/pull/471).

The amount of files is because `project` is associated to several entities, so that a replace of `project` to `repository` is in most files. Furthermore, this factory is used in almost every tests. Mainly changes are just change the name for the newer one.

⚠️ To avoid looking at all files, the important ones to check are migrations.

Next steps includes:

- Refactor `UsersProject` to `UsersRepositories`.
- Refactor `CodeOwnerProject` to `CodeOwnerRepository`
- Refactor `CodeClimateProjectMetric` to `CodeClimateRepositoryMetric`
- Refactor `ProjectDistributionData` to `RepositoryDistributionData`
- Refactor some files views (e.g. `list-project` to `repository-project`)

Resolves in part [EM-225](https://rootstrap.atlassian.net/browse/EM-225): Refactor: Refactor: Rename project entity.

PR #476, opened 2021-09-07
Refactor Project to Repository - Part 3

enzofab91:
## What does this PR do?
This PR is the third and last part of the full feature. The second part is [here](https://github.com/rootstrap/rs-code-review-metrics/pull/473).

The amount of files is because project is associated to several entities, so that a replace of project to repository is in most files. Furthermore, this factory is used in almost every tests. Mainly changes are just change the name for the newer one.

⚠️ To avoid looking at all files, the important ones to check are migrations.

Note: Jira Project entity has not been renamed. I believe is fine to keep it like now.

Resolves in part [EM-225](https://rootstrap.atlassian.net/browse/EM-225): Refactor: Refactor: Rename project entity.


santib (commented):
I think in the README there are some things that need changed like `enabled_project_codeowners_section` and  `enabled_project_per_user_graph` , right?

santib (approved):
great job 👏

lauperalti (approved):
Nice! 💯

PR #477, opened 2021-09-07
Fix/argument error

lauperalti:
## What does this PR do?
Fix bugs in background Job JiraDefectMetricUpdaterJob

Resolves [#EM-244](https://rootstrap.atlassian.net/browse/EM-244)
Resolves [#EM-249](https://rootstrap.atlassian.net/browse/EM-249)

### DIFF
--- a/app/services/jira_client/repository.rb
+++ b/app/services/jira_client/repository.rb
@@ -45,7 +45,7 @@ module JiraClient
 
     def create_request(issue_type)
       @request_params = {
-        jql: "project=#{@jira_board.jira_project_key} AND #{issue_type}",
+        jql: "project='#{@jira_board.jira_project_key}' AND #{issue_type}",
         fields: "#{JIRA_ENVIRONMENT_FIELD},created,resolutiondate,issuetype",
         expand: 'changelog'
       }
--- a/app/services/processors/jira_project_defect_escape_rate_updater.rb
+++ b/app/services/processors/jira_project_defect_escape_rate_updater.rb
@@ -33,15 +33,24 @@ module Processors
     end
 
     def issue_update!(issue, bug_fields, histories)
-      environment_field = bug_fields[JIRA_ENVIRONMENT_FIELD]
+      environment_field = parse_by_envirorment(bug_fields[JIRA_ENVIRONMENT_FIELD])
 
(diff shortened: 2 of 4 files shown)


app/services/processors/jira_project_defect_escape_rate_updater.rb
```diff
@@ -33,17 +33,26 @@ def bugs_to_update
...
         informed_at: bug_fields[:created],
         resolved_at: bug_fields[:resolutiondate] || nil,
         in_progress_at: in_progress_at(histories) || issue.in_progress_at,
-        environment: environment_field ? environment_field.first[:value].downcase : nil,
+        environment: environment_field,
         issue_type: 'bug'
       )
     end
 
+    def parse_by_envirorment(environment)
```
enzofab91, line 47: What about adding a test for this environment?

app/services/jira_client/repository.rb
```diff
@@ -45,7 +45,7 @@ def sprints
 
     def create_request(issue_type)
       @request_params = {
-        jql: "project=#{@jira_board.jira_project_key} AND #{issue_type}",
+        jql: "project='#{@jira_board.jira_project_key}' AND #{issue_type}",
```
sebastiancaraballo, line 48: [String interpolation only works with double quotes.](http://ruby-for-beginners.rubymonstas.org/bonus/string_interpolation.html)

lauperalti: In this case the interpolation is with double quotes but, inside single (or double) quotes are needed around the key string literal as part of the request params.
So, reading the documentation it may seem that it is better to escape` "...\"#{key}\"..."` ... or what about the [percent notation](https://blog.appsignal.com/2016/12/21/ruby-magic-escaping-in-ruby.html) `%(..."#{key}"...)` ? 

WDYT?

pd: i have tried all 3 options and they all worked.

enzofab91 (approved):
Nice!

app/services/jira_client/repository.rb
sebastiancaraballo: Oh, I see. Thanks for the clarification!

sebastiancaraballo (approved):

PR #480, opened 2021-09-10
Add meta tags headers for link unfurling

enzofab91:
## What does this PR do?
This PR adds meta tags for better link unfurling. In this first version, title and description are customized for each page. But at the moment, a preview image can be seen only in product metrics page.

The chart image generation is done with [Quickchart](https://quickchart.io/) API. This library generate an image from a javascript chart like the ones generated in the views with Chartkick.

**Here are some examples previews:**

When sharing product dashboard:

Screen Shot 2021-09-10 at 16 45 19

When sharing defect escape rate metric detail:

Screen Shot 2021-09-10 at 16 50 30

When sharing development cycle metric detail:

Screen Shot 2021-09-10 at 17 23 40

When sharing planned to done ratio metric detail:

Screen Shot 2021-09-10 at 17 01 37

## Notes:
Create the following env variable after deploying `QUICKCHART_URL` with value https://quickchart.io/chart/create.

Resolves #[EM-216](https://rootstrap.atlassian.net/browse/EM-216): Add OG meta tags for better link unfurling.

### DIFF
--- a/app/helpers/meta_tags_helper.rb
+++ b/app/helpers/meta_tags_helper.rb
@@ -1,0 +1,186 @@
+module MetaTagsHelper
+  def meta_title
+    title_for_metrics
+  end
+
+  def meta_description
+    description_for_metrics
+  end
+
+  def meta_image
+    generate_url_image if current_page?(products_metrics_development_metrics_path)
+  end
+
+  private
+
+  def title_for_metrics
+    if current_page?(products_development_metrics_path)
+      product_name.present? ? "#{product_name} summary" : default_description
+    elsif current_page?(products_metrics_development_metrics_path)
+      product_name + ' - ' + MetricDefinition.where(code: @query_metric_name).first&.name
+    else
+      title_for_pages
+    end
+  end
+
+  def title_for_pages
+    if current_page?(repositories_development_metrics_path)
+      "#{@repository[:name]} summary"
+    elsif current_page?(open_source_index_path)
+      'Open Source repositories'
+    elsif current_page?(departments_development_metrics_path)
+      @department.present? ? "#{@department[:name]} department metrics" : 'Departments'
+    elsif current_page?(tech_blog_path)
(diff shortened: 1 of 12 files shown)


app/helpers/meta_tags_helper.rb
```diff
@@ -0,0 +1,178 @@
...
+    generate_url_image if current_page?(products_metrics_development_metrics_path)
+  end
+
+  private
+
+  def title_for_pages
+    if current_page?(repositories_development_metrics_path)
+      "#{@repository[:name]} summary"
+    elsif current_page?(open_source_index_path)
+      'Open Source repositories'
```
lauperalti, line 34: What do you think about using `I18n` for plain text occurrences such as titles and descriptions?

enzofab91: Done!

app/services/builders/chartkick/generate_chart_image.rb
```diff
@@ -0,0 +1,119 @@
+module Builders
+  module Chartkick
+    class GenerateChartImage
+      QUICKCHART_URL = 'https://quickchart.io/chart/create'.freeze
```
lauperalti, line 4: Maybe be in  `.ENV`?

lauperalti (approved):
Very nice job! 👏

app/services/builders/chartkick/generate_chart_image.rb
enzofab91: Done!

lauperalti: Remember to add it to `.env.example`

sebastiancaraballo (approved):
Good. However, I feel `app/helpers/meta_tags_helper.rb` class should be refactored. Try calling methods dynamically instead of having `if/else` blocks. Also, analyze whether it's a good idea to split this class or not.

PR #483, opened 2021-09-16
Remove dependant destroy to repository events associations

enzofab91:
## What does this PR do?
This PR removes the `:dependant` option to the association `events` in the `repository` model, because when deleting a repository from the admin page it ends in a timeout. This is due to all the events associated to a repository that needs to be updated.

Resolves [#EM-254](https://rootstrap.atlassian.net/browse/EM-254): Keep events associations when deleting repositories.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -37,18 +37,15 @@ class Repository < ApplicationRecord
   belongs_to :language
   belongs_to :product, optional: true
 
-  has_many :events, dependent: :destroy
+  has_many :events
   has_many :repositories,
            class_name: 'Events::Repository',
-           dependent: :destroy,
            inverse_of: :repository
   has_many :pull_requests,
            class_name: 'Events::PullRequest',
-           dependent: :destroy,
            inverse_of: :repository
   has_many :reviews,
            class_name: 'Events::Review',
-           dependent: :destroy,
            inverse_of: :repository
   has_many :review_requests,
            dependent: :destroy,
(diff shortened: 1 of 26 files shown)


sebastiancaraballo (approved):

app/models/repository.rb
```diff
@@ -37,18 +37,15 @@ class Repository < ApplicationRecord
   belongs_to :language
   belongs_to :product, optional: true
 
-  has_many :events, dependent: :destroy
+  has_many :events
```
santib, line 40: should we also remove `acts_as_paranoid` from all the events models?

spec/requests/admin/repositories.rb
```diff
@@ -0,0 +1,93 @@
+require 'rails_helper'
+
+describe 'Admin::Repositories', type: :request do
```
santib, line 3: I think this should be a model test (unit) instead of a request test (integration). We just want to test that when calling `repository.destroy` the events aren't destroyed, there is no need for a request or anything else

enzofab91: I'm doing the unit test. You think is worth it to leave this test since repositories will be mainly deleted from the admin or isn't necessary?

santib: I think the changes on this PR are only related to the model, there isn't anything special in the admin or controller

santib (approved):
Great job 👏

PR #497, opened 2021-12-29
Feature/add dates from to

brunoparma88:
Change weekly range to "From To"
Add Alerts bootstrap


app/controllers/development_metrics_controller.rb
```diff
@@ -167,12 +168,23 @@ def code_climate_repository_summary
...
     )
   end
 
   def department_overview
-    Builders::DepartmentOverview.call(department, from: metric_params[:period].to_i.weeks.ago)
+    Builders::DepartmentOverview.call(department, from: @from, to: @to)
+  end
+
+  def validate_dates
+    @from = metric_params[:from]
```
santib, line 182: minor comment, but I don't like having any kind of logic on the controllers (for maintainability). So maybe you can have a `DateValidator`, `DateValidatorService`, or something like that that performs the validations. That approach would also help with duplication. I know it might seem an overkill for just 4 lines, so yeah, up to you

app/helpers/dropdown_selection_helper.rb
```diff
@@ -6,4 +6,20 @@ def chosen_user
...
+  def chosen_to
+    params.dig(:metric, :to) || initial_to
+  end
+
+  def initial_to
+    Time.zone.now.strftime('%Y-%m-%d')
+  end
+
+  def initial_from
+    (Time.zone.now - 28.days).strftime('%Y-%m-%d')
```
santib, line 23:
```suggestion
    28.days.ago.strftime('%Y-%m-%d')
```

BTW why 28? I guess `2.weeks` might be a better default? IDK

app/helpers/dropdown_selection_helper.rb
```diff
@@ -6,4 +6,20 @@ def chosen_user
   def chosen_period
```
santib, line 6: are these `period` methods dead? I mean, can we remove them or are they still in use?

app/services/metrics/group/weekly.rb
```diff
@@ -15,7 +16,10 @@ def interval
       end
 
       def value_timestamp
-        (current_time - @prev.weeks).beginning_of_week..current_time.end_of_week
+        now = Time.zone.now
+        from = @from || (now - 28.days).strftime('%Y-%m-%d')
```
santib, line 20: same comment

app/views/development_metrics/departments.erb
```diff
@@ -1,5 +1,5 @@
 <%= render 'development_metrics/departments_dropdown' %>
-<%= render 'development_metrics/development_metrics_sidebar', enabled_users_section: @enabled_users_section %>
+<%= render 'development_metrics/development_metrics_sidebar', enabled_users_section: @enabled_users_section  %>
```
santib, line 2:
```suggestion
<%= render 'development_metrics/development_metrics_sidebar', enabled_users_section: @enabled_users_section %>
```

app/views/shared/_alert.erb
```diff
@@ -0,0 +1,6 @@
+<% flash.each do |key, message| %>
+<div class="metrics-alert alert alert-warning alert-dismissible fade show" role="alert">
+    <%=  message %>
+  <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
+</div>
+<% end %>
```
santib, line 6:
```suggestion
<% flash.each do |key, message| %>
  <div class="metrics-alert alert alert-warning alert-dismissible fade show" role="alert">
    <%=  message %>
    <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
  </div>
<% end %>
```

app/views/shared/_alert.erb
```diff
@@ -0,0 +1,6 @@
+<% flash.each do |key, message| %>
+<div class="metrics-alert alert alert-warning alert-dismissible fade show" role="alert">
```
santib, line 2: Why `alert-warning` is hardcoded? the flash can be a warning, a notice, an error, etc, right?

app/controllers/development_metrics_controller.rb
```diff
@@ -167,12 +168,23 @@ def code_climate_repository_summary
...
       technologies: []
     )
   end
 
   def department_overview
-    Builders::DepartmentOverview.call(department, from: metric_params[:period].to_i.weeks.ago)
+    Builders::DepartmentOverview.call(department, from: @from, to: @to)
+  end
+
+  def validate_dates
```
santib, line 181: Is there any spec testing the validation?

santib (approved):
Good job!

app/views/shared/_alert.erb
brunoparma88: In this app only need one warning.

martinjaimem (approved):

PR #508, opened 2022-02-10
add value on PR details

brunoparma88:
Add hours to each PR.

image


spec/requests/pull_requests/time_to_merge_prs/index_spec.rb
```diff
@@ -89,7 +89,7 @@
 
     it 'returns data only for all repositories' do
       assigns(:pull_requests).each do |_, pull_request_url|
-        expect(all_pull_requests_url).to include(pull_request_url.first)
+        expect(all_pull_requests_url).to include(pull_request_url.first[:html_url])
```
megatux, line 92: what about validating the new _value_ added?

megatux (commented):
LGTM. Added a comment on tests

app/assets/stylesheets/application.scss
```diff
@@ -143,6 +143,10 @@ main {
   background-color: #d2f4ea;
 }
 
+.hours-pr{
+  color: grey
```
megatux, line 147: missing semicolon

spec/requests/pull_requests/time_to_merge_prs/index_spec.rb
brunoparma88: The value already existed. It is validated in "Events PullRequest Spec"

yarn.lock
```diff
@@ -1746,9 +1746,9 @@ caniuse-api@^3.0.0:
     lodash.uniq "^4.5.0"
 
 caniuse-lite@^1.0.0, caniuse-lite@^1.0.30000981, caniuse-lite@^1.0.30001043, caniuse-lite@^1.0.30001061:
```
enzofab91, line 1748: I think this file is not from this PR. WDYT?

brunoparma88: I separate it for another PR

santib (approved):
I'm not sure why we don't use feature/system specs and check that the correct texts are being rendered, but 🤷‍♂️

PR #513, opened 2022-02-18
change weeks group by

brunoparma88:
fix defect escape rate bug when date is greater than one year

### DIFF
--- a/app/services/metrics/base.rb
+++ b/app/services/metrics/base.rb
@@ -16,7 +16,7 @@ module Metrics
     def week_intervals
       from = metric_interval.first.to_date
       to = metric_interval.last.to_date
-      (from..to).group_by(&:cweek)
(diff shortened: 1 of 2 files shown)


app/services/metrics/base.rb
```diff
@@ -16,7 +16,7 @@ def call
     def week_intervals
       from = metric_interval.first.to_date
       to = metric_interval.last.to_date
-      (from..to).group_by(&:cweek)
+      (from..to).group_by { |t| t.strftime("%W/%Y")}
```
sebastiancaraballo, line 19: Please add test

brunoparma88: the test was added

sebastiancaraballo (approved):

santib (approved):

PR #525, opened 2022-03-04
minimal changes on styles

brunoparma88:
*Fix padding from welcome and sidebar
*Change style html to bootstrap class
*Change overflow-y: auto on sidebar

image


lauperalti (approved):
🙌

PR #532, opened 2022-04-13
add enabled to product

brunoparma88:
EM-323(https://rootstrap.atlassian.net/browse/EM-323)
Create enabled flag to projects


app/admin/products.rb
```diff
@@ -1,13 +1,14 @@
 ActiveAdmin.register Product do
   permit_params :id, :description, :name, :jira_key, :new_jira_key,
-                jira_board_attributes: %i[id jira_project_key],
-                repository_ids: []
+                :enabled, jira_board_attributes: %i[id jira_project_key],
+                          repository_ids: []
```
sebastiancaraballo, line 4: Is alignment okay?

db/migrate/20220412181602_add_enabled_to_products.rb
```diff
@@ -0,0 +1,7 @@
+class AddEnabledToProjects < ActiveRecord::Migration[6.0]
+  def change
+    change_table :products do |t|
+      t.boolean :enabled, null: false, default: true
```
sebastiancaraballo, line 4: Maybe it could be worth adding an index

app/models/product.rb
```diff
@@ -24,7 +25,7 @@ class Product < ApplicationRecord
   has_many :metrics, as: :ownable, dependent: :destroy
 
   validates :name, presence: true, uniqueness: true
-
+  validates :enabled, presence: true
```
sebastiancaraballo, line 28: Would it be a good idea to set `enabled` at the default_scope? Just asking. https://apidock.com/rails/ActiveRecord/Base/default_scope/class

app/admin/products.rb
brunoparma88: Rubocop put it there.

app/models/product.rb
brunoparma88: I didn't know about default_scope, it's a good idea.

app/models/product.rb
```diff
@@ -5,26 +5,28 @@
...
 #
 #  index_products_on_deleted_at  (deleted_at)
+#  index_products_on_enabled     (enabled)
 #  index_products_on_name        (name)
 #
 
 class Product < ApplicationRecord
   acts_as_paranoid
-
+  scope :enabled, lambda {
```
sebastiancaraballo, line 22: This should go at the end. Take a look at [this](https://github.com/rootstrap/tech-guides/blob/master/ruby/rails.md#macro-style-methods)

sebastiancaraballo (approved):

app/models/product.rb
```diff
@@ -5,26 +5,28 @@
 #  id          :bigint           not null, primary key
 #  deleted_at  :datetime
 #  description :string
+#  enabled     :boolean          default(TRUE), not null
```
lauperalti, line 8: Could we add presence validation of the enabled attribute in the model and in spec?

app/models/alert.rb
```diff
@@ -3,7 +3,7 @@
 # Table name: alerts
 #
 #  id             :bigint           not null, primary key
-#  active         :boolean
+#  active         :boolean          default(FALSE), not null
```
lauperalti, line 6: Could we add presence validation of the `active` attribute in the model and in spec?

lauperalti (approved):
LGTM. Left some minor comments. 🚀

app/models/alert.rb
brunoparma88: the field "activate" doesn't change in this PR. Maybe annotate was not run in the previous PR

PR #536, opened 2022-05-05
fix/see-details-pr-size-view-error

FernandoCiballero:
## What does this PR do?
The intention of this fix is to avoid that, when listing the detail of PRs in the `PR size` section, it does not include PRs that are `Closed`.

Screen Shot 2022-05-05 at 17 03 50

Resolves #XX

### DIFF
--- a/app/services/builders/distribution/pull_requests/pull_request_size_repository.rb
+++ b/app/services/builders/distribution/pull_requests/pull_request_size_repository.rb
@@ -23,7 +23,10 @@ module Builders
           @pr_sizes ||= ::Events::PullRequest.where(created_at: @from..@to)
                                              .joins(:repository)
                                              .where(repositories: { name: @repository_name })
-                                             .where.not(events_pull_requests: { html_url: nil })
+                                             .where.not(
+                                               events_pull_requests: { html_url: nil },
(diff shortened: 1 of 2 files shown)


femartinez (approved):
LGTM

app/services/builders/distribution/pull_requests/pull_request_size_repository.rb
```diff
@@ -23,7 +23,10 @@ def pr_sizes
           @pr_sizes ||= ::Events::PullRequest.where(created_at: @from..@to)
                                              .joins(:repository)
                                              .where(repositories: { name: @repository_name })
-                                             .where.not(events_pull_requests: { html_url: nil })
+                                             .where.not(
+                                               events_pull_requests: { html_url: nil },
+                                               size: nil
+                                             )
```
santib, line 29: can we add a test for this?

santib: I'd like to have tested the following cases:
- html_url = nil and size = nil
- html_url not nil and size nil
- html_url nil and size not nil
- html_url not nil and size not nil

lauperalti (approved):

PR #561, opened 2022-12-15
[EM-353] added presenter for repo metrics

esparratacus:
## What does this PR do?
Creates a presenter taking care of the logic displaying chart data

Resolves [EM-353](https://rootstrap.atlassian.net/browse/EM-353)

### DIFF
--- a/app/presenters/repository_metric_presenter.rb
+++ b/app/presenters/repository_metric_presenter.rb
@@ -1,0 +1,23 @@
+class RepositoryMetricPresenter
+
+  METRIC_FILTERS = [:per_repository,
+                    :per_users_repository,
+                    :per_repository_distribution]
+
+
+  METRIC_FILTERS.each do |filter|
+    define_method("#{filter.to_s}_has_data_to_display?") do
+      content = @metric.dig(filter, 0, :data)
+      !(content.nil? || content.empty?)
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


app/presenters/repository_metric_presenter.rb
```diff
@@ -0,0 +1,17 @@
+class RepositoryMetricPresenter
+  METRIC_FILTERS = %i[per_repository per_users_repository per_repository_distribution].freeze
+  METRIC_FILTERS.each do |filter|
+    define_method("#{filter}_has_data_to_display?") do
+      content = @metric.dig(filter, 0, :data)
```
santib, line 5: I think you can use the `metric` `attr_reader` here

app/presenters/repository_metric_presenter.rb
```diff
@@ -0,0 +1,17 @@
...
+      content.present?
+    end
+  end
+
+  attr_reader :metric, :metric_name, :metric_explanation
+
+  def initialize(metric_hash, metric_definition)
+    @metric_name = metric_definition.name
```
santib, line 13: does `metric_definition.name` work? because in the view I see we were doing `metric_definition[:name]`

santib (approved):
Good job 👍

brunoparma88 (approved):

PR #563, opened 2022-12-20
Fix see details bug

esparratacus:
## What does this PR do?

bug fix: Remove see details button when there is no data to display on a metric


santib (approved):
why are there 63 commits? something feels off regarding git..

brunoparma88 (commented):
I think if you do a fix you should do it in a separate PR. And then merge develop a master in another.

brunoparma88 (approved):

PR #565, opened 2022-12-27
Project name bug

esparratacus:
## What does this PR do?

resolves project name with spaces bug


santib (approved):
Please when doing these PRs from `develop` to `master` name them something like `Release x` so it's clear that all the code here was already reviewed

brunoparma88 (approved):

PR #576, opened 2024-08-02
fix external pull requests processor job error

rodrieiz:
## What does this PR do?

This pr is to resolve some job issues:
- External pull requests processor job:
This job generates GitHub URLs with the username parameter of the employees. The problem is that the username comes from our database and in some cases, they are outdated, meaning the person has changed their username and the URL is no longer valid (several cases involve people who no longer work at the company).

image
image

- JiraDefectMetricsUpdaterJob
This job also is generating a url for jira sprints that raise not found error.
image

image


enzofab91 (approved):
I'm not sure if `gemfile.lock` is needed here, if you update some gem the `gemfile` should be in this PR too right?

Alxzu (approved):
We can leave the `gemfile.lock` for another PR

PR #580, opened 2024-08-29
[EM-381] Add unique index on repository table

rodrieiz:
[EM-381](https://rootstrap.atlassian.net/browse/EM-381?atlOrigin=eyJpIjoiMWRhOWQ5ZjMyYzgxNDY0NThiYzYxOGYyNWVjYTFiODgiLCJwIjoiaiJ9)

## What does this PR do?
We identified that there are a couple of repositories in the database with duplicate github_ids, and the job that updates the repositories fails in those cases because when it tries to update the data using find_or_initialize_by(:github_id) and then calls .save, it fails due to a uniqueness validation in the model (but there are two records in the database with the same github_id).

Resolves #XX

[EM-381]: https://rootstrap.atlassian.net/browse/EM-381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/structure.sql
```diff
@@ -9,6 +9,20 @@ SET xmloption = content;
...
+CREATE EXTENSION IF NOT EXISTS pg_stat_statements WITH SCHEMA public;
+
+
+--
+-- Name: EXTENSION pg_stat_statements; Type: COMMENT; Schema: -; Owner: -
+--
+
+COMMENT ON EXTENSION pg_stat_statements IS 'track planning and execution statistics of all SQL statements executed';
+
+
```
enzofab91, line 25: This don't seems related to the changes but with postgresql version 🤔

enzofab91 (approved):

db/structure.sql
enzofab91: I don't know, I'm fine to keep it if your version is same as production version.

rodrieiz: yes both are 16.x

claramaffei (approved):

PR #601, opened 2025-05-19
Add review coverage model

gonzaM:
## What does this PR do?
This PR introduces a new code review metric that measures the percentage of files in a pull request that have received comments during the review process.

#### Changes

- Add ReviewCoverage model with associations and validations
- Create necessary database migrations and schema updates
- Extend GithubClient to fetch PR comments
- Modify PullRequest model to build review coverage metrics upon merge
- Add new tests and adjust existent ones.
- Implement backfill rake task for existing merged PRs

### DIFF
--- a/app/models/review_coverage.rb
+++ b/app/models/review_coverage.rb
@@ -22,18 +22,19 @@
 
 class ReviewCoverage < ApplicationRecord
   acts_as_paranoid
-  
+
   belongs_to :pull_request, class_name: 'Events::PullRequest'
-  
+
   validates :total_files_changed, :files_with_comments_count, presence: true
   validates :pull_request_id, uniqueness: true
-  validates :total_files_changed, :files_with_comments_count, numericality: { greater_than_or_equal_to: 0 }
-  
+  validates :total_files_changed, :files_with_comments_count,
+            numericality: { greater_than_or_equal_to: 0 }
+
   before_save :calculate_coverage_percentage
-  
+
   private
-  
+
   def calculate_coverage_percentage
     self.coverage_percentage = (files_with_comments_count.to_f / total_files_changed).round(2)
   end
-end 
+end
--- a/app/models/review_coverage.rb
+++ b/app/models/review_coverage.rb
@@ -1,0 +1,39 @@
+# == Schema Information
+#
+# Table name: review_coverages
+#
+#  id                        :bigint           not null, primary key
+#  coverage_percentage       :decimal(, )      not null
+#  deleted_at                :datetime
+#  files_with_comments_count :integer          not null
+#  total_files_changed       :integer          not null
+#  created_at                :datetime         not null
+#  updated_at                :datetime         not null
+#  pull_request_id           :bigint           not null
+#
+# Indexes
+#
+#  index_review_coverages_on_pull_request_id  (pull_request_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (pull_request_id => events_pull_requests.id)
+#
+
+class ReviewCoverage < ApplicationRecord
+  acts_as_paranoid
+  
+  belongs_to :pull_request, class_name: 'Events::PullRequest'
+  
+  validates :total_files_changed, :files_with_comments_count, presence: true
+  validates :pull_request_id, uniqueness: true
+  validates :total_files_changed, :files_with_comments_count, numericality: { greater_than_or_equal_to: 0 }
+  
+  before_save :calculate_coverage_percentage
+  
+  private
+  
+  def calculate_coverage_percentage
+    self.coverage_percentage = (files_with_comments_count.to_f / total_files_changed).round(2)
+  end
+end 
--- a/app/models/review_coverage.rb
+++ b/app/models/review_coverage.rb
@@ -29,12 +29,4 @@ class ReviewCoverage < ApplicationRecord
   validates :pull_request_id, uniqueness: true
   validates :total_files_changed, :files_with_comments_count,
             numericality: { greater_than_or_equal_to: 0 }
-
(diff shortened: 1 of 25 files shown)


app/models/review_coverage.rb
```diff
@@ -0,0 +1,40 @@
...
+  acts_as_paranoid
+
+  belongs_to :pull_request, class_name: 'Events::PullRequest'
+
+  validates :total_files_changed, :files_with_comments_count, presence: true
+  validates :pull_request_id, uniqueness: true
+  validates :total_files_changed, :files_with_comments_count,
+            numericality: { greater_than_or_equal_to: 0 }
+
+  before_save :calculate_coverage_percentage
```
rodrieiz: it's necessary to be callback or can be calculated when creating/updating the object? Just asking because I think is a good practice to avoid this type of callbacks

app/services/github_client/pull_request.rb
```diff
@@ -18,6 +18,11 @@ def files
       get_all_paginated_items(url, MAX_FILES_PER_PAGE)
     end
 
+    def comments
+      url = "repos/#{repository.full_name}/pulls/#{pull_request.number}/comments"
+      get_all_paginated_items(url, MAX_FILES_PER_PAGE)
```
rodrieiz, line 26: mm should we handle or log possible errors here?

spec/factories/events/pull_requests.rb
```diff
@@ -53,5 +53,26 @@
...
+        create(
+          :merge_time,
+          pull_request: pull_request,
+          value: evaluator.merge_time
+        )
+        create(
+          :review_coverage,
+          pull_request: pull_request
+        )
+      end
```
rodrieiz, line 75: I think this should be created manually in each test that requires it. Maybe using a trait isn’t too bad, as long as we don’t overuse it, wdyt?

spec/tasks/review_coverage_backfill_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+require 'rake'
+
+# rubocop:disable RSpec/DescribeClass
```
rodrieiz, line 4: I way to find to avoid this is the task only calls another file  review_coverage.rb
 that have all the logic. feel free to ignore

db/migrate/20250430151601_add_review_coverage_to_metric_name_type.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def down
+    execute <<-SQL
+      ALTER TYPE metric_name RENAME TO metric_name_old;
+      CREATE TYPE metric_name AS ENUM ('review_turnaround', 'blog_visits', 'merge_time', 'blog_post_count', 'open_source_visits', 'defect_escape_rate', 'pull_request_size', 'development_cycle');
+      ALTER TABLE metrics ALTER COLUMN name TYPE metric_name USING name::text::metric_name;
+      DROP TYPE metric_name_old;
+    SQL
+  end
+end
```
rodrieiz, line 18: this looks very strange to me. How this enum defined ? is not enough to add the new enum options in the model?

db/migrate/20250430151601_add_review_coverage_to_metric_name_type.rb
```diff
@@ -0,0 +1,18 @@
...
+      ALTER TYPE metric_name ADD VALUE 'review_coverage';
+    SQL
+  end
+
+  def down
+    execute <<-SQL
+      ALTER TYPE metric_name RENAME TO metric_name_old;
+      CREATE TYPE metric_name AS ENUM ('review_turnaround', 'blog_visits', 'merge_time', 'blog_post_count', 'open_source_visits', 'defect_escape_rate', 'pull_request_size', 'development_cycle');
+      ALTER TABLE metrics ALTER COLUMN name TYPE metric_name USING name::text::metric_name;
+      DROP TYPE metric_name_old;
```
santib, line 15: aren't we losing all the old values?

santib (approved):

db/migrate/20250430151601_add_review_coverage_to_metric_name_type.rb
gonzaM: This enum is defined as a type at a [database level](https://www.postgresql.org/docs/current/datatype-enum.html), it's not an integer, and that's why I need to add the value with a migration.

db/migrate/20250430151601_add_review_coverage_to_metric_name_type.rb
gonzaM: I understand the previous line is transferring the values.

spec/factories/events/pull_requests.rb
gonzaM: Given that this logic is used in many tests, I thought it was a good idea to have the trait to avoid repeating all the logic. Also, it's a simple logic and not a complex trait. Do you think it's ok in this case?

app/services/github_client/pull_request.rb
```diff
@@ -16,10 +16,29 @@ def get
...
+      handle_exception(exception)
+      []
     end
 
     private
 
     attr_reader :pull_request
+
+    def handle_exception(exception)
+      Honeybadger.notify(exception)
```
gonzaM, line 37: I added Honeybadger because it is the tool being used, but I don't have access to it. I think in a future PR, we can move to Sentry or something similar. I added it anyway to keep track of this error notification.

app/models/review_coverage.rb
gonzaM: Moved! 💪

rodrieiz (approved):
looks great!

PR #609, opened 2025-09-29
Add zero comments percetage

gonzaM:
### Add percentage of PRs with zero comments to Review Coverage

This PR enhances Review Coverage by introducing a new metric: the percentage of PRs with zero comments.

### Motivation
The current review coverage average number provides limited value, as it does not clearly show how much reviewing is actually happening. To make this more actionable, we add the percentage of PRs with zero comments, emphasizing that what matters is ensuring each PR receives some level of review, rather than the total number of comments, which does not directly reflect quality. 

### Summary
- Adds zero-coverage percentage to Review Coverage (percent of PRs with 0 comments).
- Unifies distribution payload by nesting computed values under `interval_metrics`.
- Updates UI to display new metric and use unified structure.
- Adjusts services, controller, and specs accordingly.

### Changes
- Data structure refactoring: Consolidates metric calculations under interval_metrics instead of separate top-level keys
- New metric calculation: Adds percentage of PRs with zero comments to Review Coverage
- UI updates: Displays new zero-coverage percentage badge and migrates existing badges to use unified structure.

### Note
The "department metrics" are no longer used, which is why the "success rate" was not updated and the test was removed. In a future PR, the plan is to remove the "Department" logic altogether to simplify the code.

Screenshot 2025-09-29 at 3 44 18 PM

### DIFF
--- a/app/services/builders/chartkick/repository_distribution_data.rb
+++ b/app/services/builders/chartkick/repository_distribution_data.rb
@@ -44,7 +44,7 @@ module Builders
       end
 
       def total_records
-        @total_records ||= retrieve_records.count
+        @total_records ||= retrieve_records.size
       end
 
       def success_rate
--- a/app/services/builders/chartkick/repository_distribution_data.rb
+++ b/app/services/builders/chartkick/repository_distribution_data.rb
@@ -2,13 +2,10 @@ module Builders
   module Chartkick
     class RepositoryDistributionData < Builders::Chartkick::Base
       def call
-        intervals = build_distribution_data(retrieve_records)
-
         [{
           name: repository_name,
           data: intervals,
-          success_rate: build_success_rate(repository_name, metric_name, intervals),
-          avg: build_average
+          interval_metrics: build_interval_metrics
         }]
       end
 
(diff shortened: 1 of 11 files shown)


app/services/builders/chartkick/repository_distribution_data.rb
```diff
@@ -34,20 +31,41 @@ def metric
...
       end
 
-      def build_average
-        return if retrieve_records.empty?
+      def total_value
+        @total_value ||= retrieve_records.sum { |record| metric.value_for_average(record) }
+      end
+
+      def total_records
+        @total_records ||= retrieve_records.count
```
santib:
```suggestion
        @total_records ||= retrieve_records.size
```

so that if records are already in memory it'll save an extra DB query

app/services/builders/chartkick/repository_distribution_data_metrics/review_coverage.rb
```diff
@@ -22,6 +22,15 @@ def resolve_interval(entity)
         def value_for_average(entity)
           entity.coverage_percentage * 100
         end
+
+        def zero_coverage_percentage(records)
+          return 0 if records.empty?
+
+          zero_coverage_count = records.with_zero_coverage.count
+          total_count = records.count
```
santib:
```suggestion
          total_count = records.count
          return 0 if total_count.zero?

          zero_coverage_count = records.with_zero_coverage.count
```

santib (approved):
👮 

left a couple of comments so this review isn't considered a 0comments review 🤣

